PR #24, opened 2026-04-15
fix(p1): encode VFO frequency as raw Hz — Phase 3G-11

boydsoftprez:
## Summary

- Restores VFO tuning on Protocol 1 hardware. `P1RadioConnection::composeCcBankRxFreq` and `composeCcBankTxFreq` were pre-converting Hz to an NCO phase word; P1 firmware expects raw Hz.
- First entry in new **Phase 3G-11 (P1 Field-Report Fixes)** bucket. Other in-flight P1 alpha-tester fixes will land as additional bullets under the same phase.

## Root cause

Tracked down from alpha tester pcap4 (2026-04-15, ANAN-10E). 319 consecutive `C0=0x04` Metis frames over 20 seconds all pinned at phase word `0x080D5555`, producing an aliased waterfall and a VFO that would not track the dial.

Thetis `NetworkIO.cs:215-223 VFOfreq()` selects the encoding on `CurrentRadioProtocol`:

```csharp
if (CurrentRadioProtocol == RadioProtocol.USB)     // USB == P1  (cmaster.cs:520)
    SetVFOfreq(id, f_freq, tx);                    // raw Hz
else SetVFOfreq(id, Freq2PhaseWord(f_freq), tx);   // phase word (P2)
```

The native `networkproto1.c:476-494` then splats `prn->{tx,rx}[0].frequency` directly into C1..C4 big-endian with no conversion. NereusSDR was running `Freq2PhaseWord` unconditionally; the in-file comment misread the Thetis `else` branch as the common path.

## Fix

- `composeCcBankRxFreq` / `composeCcBankTxFreq`: splat `freqHz` big-endian into C1..C4 as raw Hz, cast to `quint32`.
- Header comments rewritten to cite `NetworkIO.cs:215-223` and `networkproto1.c:476-494` correctly and record the pcap4 root cause.
- `docs/MASTER-PLAN.md`: new Phase 3G-11 bucket.

## Test plan

- [x] `tst_p1_wire_format` — **29 passed, 0 failed**. The `rxFrequencyEncodedBigEndian32` / `txFrequencyEncodedBigEndian32` cases already asserted raw Hz; the implementation had silently drifted past them. Fix realigns code with the pre-existing assertions.
- [x] Full NereusSDR.exe build succeeds on mingw1310_64 + Qt 6.11.0.
- [x] Local smoke test on ANAN-G2 (Saturn / P2) — connect, stream, clean shutdown (exercises unrelated P2 path; sanity only).
- [x] **Verified working on tester's live ANAN-10E (Protocol 1) 2026-04-15** — VFO now tracks the dial.
- [ ] CI `tst_p1_wire_format` green.

## Follow-ups (out of scope)

- `NEREUS_BUILD_TESTS` suite does not appear to be wired into `ci.yml` — `tst_p1_wire_format` had correct assertions but the implementation drifted undetected. Worth a separate task to ensure opt-in tests actually run on PR.
- Possible separate RX1 signal-chain staleness: in the same pcap, RX1 bytes were byte-identical across all 319 frames while TX varied. May still be an issue post-fix; waiting on follow-up pcap to confirm. Would belong under 3G-11.

### DIFF
--- a/docs/MASTER-PLAN.md
+++ b/docs/MASTER-PLAN.md
@@ -508,6 +508,11 @@ Shipped as three sequential PRs off `main`. Strict dependency: PR1 helpers feed 
 
 **Non-goals:** RX2/TX display surface, Spectrum Overlay flyout refactors, skin system, Thetis default-value adoption beyond the seven PR2 recipes. Source-first protocol per CLAUDE.md governs everything else; 3G-8's §10 divergence exception is **not** extended.
 
+### Phase 3G-11: P1 Field-Report Fixes
+**Goal:** Close out the bugs surfaced by alpha testers running Protocol 1 hardware after 3I shipped. Grouped under one phase so each fix doesn't have to claim its own number. Each bullet below lands from its own session / branch.
+
+- **P1 RX/TX VFO frequency encoding** — `P1RadioConnection::composeCcBankRxFreq` / `composeCcBankTxFreq` were pre-converting Hz to an NCO phase word (`freqHz * 2^32 / 122.88e6`). P1 firmware expects raw Hz; Thetis `NetworkIO.cs:215-223` only calls `Freq2PhaseWord` on the P2 (`ETH`) branch, and the native `networkproto1.c:476-494` splats `prn->{tx,rx}[0].frequency` directly into C1..C4 with no conversion. Symptom reproduced in alpha tester pcap4 (2026-04-15, ANAN-10E): 319 consecutive `C0=0x04` frames pinned at phase word `0x080D5555`, aliased waterfall, VFO did not track dial. Fix aligns with pre-existing `tst_p1_wire_format` assertions that had silently drifted unvalidated. Branch `fix/p1-freq-encoding`, confirmed working against live ANAN-10E 2026-04-15.
+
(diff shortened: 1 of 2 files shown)


boydsoftprez:
Added doc updates in c92d69d.

While drafting the CHANGELOG I noticed `docs/architecture/radio-abstraction.md` had the P1 and P2 frequency encodings **inverted** — it claimed P1 uses a phase word and P2 uses raw Hz, which is the opposite of reality (verified against Thetis `NetworkIO.cs:215-223`, `networkproto1.c:476-494`, and the shipping `P2RadioConnection` phase-word code). That doc error is almost certainly what misled whoever wrote the broken P1 implementation in the first place, so fixing it is part of the same root-cause fix.

Changes in c92d69d:
- `docs/architecture/radio-abstraction.md` — swapped the §5 (P1) and §6 (P2) interface comments, the P1 vs P2 comparison table, the §5 C&C register map, and rewrote the "Frequency Phase Word Calculation" subsection as "Frequency Encoding (Raw Hz)" with Thetis source citations.
- `CHANGELOG.md` — new `[Unreleased]` block describing the P1 fix and the doc correction.

No code changes beyond what was already in 5dec1cb. Test suite still green.

PR #65, opened 2026-04-19
Phase 3O Sub-Phase 7: VirtualCableDetector (Windows BYO helper)

boydsoftprez:
## Summary
- Adds `src/core/audio/VirtualCableDetector.{h,cpp}` — pure-function regex
  matcher + live `PortAudioBus` device scanner for well-known virtual-audio-cable
  products. Used by the Windows "bring-your-own-cable" first-run helper (later
  sub-phase) to detect already-installed products and offer vendor install URLs
  for missing ones.
- 12 rule set (order is load-bearing): reserved `NereusSDR VAX <n>` namespace
  first, then VB-Audio `CABLE-{A..D}`, `Hi-?Fi Cable`, base `CABLE `, Muzychenko
  VAC (`Virtual Audio Cable` + `Line <n>`), VoiceMeeter, Dante, FlexRadio DAX.
- Cross-platform C++ — builds and tests on every OS. No platform guards; no
  OS-specific I/O. Only runtime dependency beyond Qt6 is the existing
  `PortAudioBus`.
- Independent of PR #64 (Sub-Phase 6 / LinuxPipeBus) — branches off the same
  `origin/main` base and can merge in either order.

## Tests
9 QtTest slots (all passing locally):
- 8 required (`detectsVbCableBase`, `detectsVbCableAB`, `detectsVac`,
  `detectsVoicemeeter`, `detectsDante`, `detectsFlexDax`,
  `detectsReservedNereusVax`, `unknownDeviceIsNone`)
- 1 regression guard (`cableAbDoesNotMatchCableA` — ensures the `^CABLE-A `
  trailing-space anchor correctly rejects `CABLE-AB`-style names)

## Test plan
- [ ] Linux CI: `tst_virtual_cable_detector` builds and runs
- [ ] macOS CI: same
- [ ] Windows CI: same
- [ ] `compliance-inventory.py` classifies the new files as
      `nereussdr-original` (no port → no attribution row expected)

## Scope notes
- `scan()` live-enumeration only wires `PortAudioBus::{hostApis,
  outputDevicesFor, inputDevicesFor}`. Manual Windows-host verification with
  real virtual cables is deferred to the first-run UI sub-phase.
- Multi-cable `channel` field in `DetectedCable` is currently always 0;
  per-channel parsing is a later enhancement.

---
J.J. Boyd ~ KG4VCF

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -331,6 +331,7 @@ set(CORE_SOURCES
     src/core/StepAttenuatorController.cpp
     src/core/audio/MasterMixer.cpp
     src/core/audio/PortAudioBus.cpp
+    src/core/audio/VirtualCableDetector.cpp
 )
 
 # CoreAudioHalBus — macOS-only IAudioBus implementation bridging to the
--- a/src/core/audio/VirtualCableDetector.cpp
+++ b/src/core/audio/VirtualCableDetector.cpp
@@ -1,0 +1,74 @@
+// src/core/audio/VirtualCableDetector.cpp (NereusSDR)
+// Implements VirtualCableDetector: regex product matching + PortAudio scan.
+// NereusSDR-original; no Thetis/AetherSDR port.
+#include "VirtualCableDetector.h"
+#include <QRegularExpression>
+#include "PortAudioBus.h"
+
+using namespace NereusSDR;
+
+VirtualCableProduct VirtualCableDetector::matchProduct(const QString& n) {
+    // Rule order is load-bearing:
+    //   - NereusSdrVax first (reserved prefix must not be shadowed by CABLE rules)
+    //   - CABLE-A/B/C/D before CABLE (or the base rule shadows the variants)
+    //   - VbHiFiCable before CABLE (name contains "Hi-Fi Cable", not "CABLE ")
+    // Rules are function-local statics — QRegularExpression constructed once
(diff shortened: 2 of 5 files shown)


boydsoftprez:
Acknowledging both P2 findings — deferring to the first-run UI sub-phase:

**1. `^Line \d+` breadth (VAC vs generic "Line N" hardware)** — real concern,
but the rule was explicitly authorized in the plan because VAC installations
that rename the default device strip the "Virtual Audio Cable" substring. The
right narrow fix (drop `^Line \d+` and rely on the `Virtual Audio Cable` rule)
loses those edge-case installs. Better to validate against actual Windows
hardware + VAC configs during the first-run UI work before tightening.

**2. `scan()` host-API deduplication** — real gap. The right policy depends on
how the first-run UI presents results: showing all 4 (MME / DirectSound /
WASAPI / WDM-KS) lets the user pick their preferred backend; collapsing to
one loses that choice. That UI call hasn't been made yet, and `DetectedCable`
lacks a `hostApi` field to support either strategy cleanly. Will revisit
alongside the UI design — at which point the struct likely grows a `hostApi`
field and `scan()` either groups or collapses per UX decision.

Both are tracked as Phase 3O Sub-Phase 9 (VAX UI) scope.

---
J.J. Boyd ~ KG4VCF

PR #160, opened 2026-05-01
fix(hl2): N2ADR per-MAC persistence + signed S-ATT spinbox range

### LINKED ISSUE #105: Step attenuator range incorrect for Hermes Lite 2
### Bug Description

The step attenuator is currently showing 0-31 for HL2 instead of the 0-60 range you mentioned.  Generally I believe this is shown as -12 to 48 for HL2 but I doubt that really matters.  There is no slider widget but when double clicked numeric values show the level and do affect the gain level when adjusted.

### Steps to Reproduce

1. Double click on step attenuator control and scroll with mouse wheel to see range move from 0 -31.

### Radio Model & Protocol

Hermes Lite 2

### Radio Firmware Version

74

### NereusSDR Version

0.2.1

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs

```text

```

boydsoftprez:
## Summary
Two HL2 bench-caught bugs running on a Hermes Lite 2 (hermes-filter-debug branch).

- **Bug 1 — S-ATT spinbox clamped at 0** despite HL2's signed `-28..+32` range
  advertised in `BoardCapabilities` (mi0bot `setup.cs:16085-16086 [v2.10.3.13-beta2]`).
  Three call-sites now propagate the board's signed range end-to-end:
  `RxApplet::setBoardCapabilities` re-ranges `m_stepAttSpin` from caps on every
  `currentRadioChanged` (was: only antenna-button visibility);
  `GeneralOptionsPage` uses `m_ctrl->minAttenuation()` (was: hardcoded `0`);
  `RadioModel::connectToRadio` pushes `caps.attenuator.{min,max}Db` into the
  controller on every connect.

- **Bug 2 — N2ADR Filter board enable lost on app restart**, with a destructive
  twist: opening `Setup → HL2 I/O` actively wiped both the persisted setting
  and the OcMatrix. Root cause was a write/read scope mismatch — write went
  to global `"hl2IoBoard/n2adrFilter"`, `HardwarePage::restoreSettings` read
  per-MAC. Empty per-MAC → restore defaulted to `false` → fired
  `onN2adrToggled(false)` → wiped global + OcMatrix every Setup-tab open.
  Fix routes the write through `settingChanged()` → `HardwarePage::wire()` →
  `setHardwareValue(mac, ...)` (the path every other Hardware sub-tab uses).
  `RadioModel::connectToRadio` reads per-MAC. `restoreSettings` is no-op when
  the key is absent. One-shot `AppSettings::migrateLegacyN2adrFilter` (called
  from `main.cpp`) copies any legacy global value into `hardware/<mac>/...` for
  every saved HL2, then deletes the global; non-HL2 saved radios skipped
  (`chkHERCULES` on a non-HERMESLITE radio is Apollo, not N2ADR — mi0bot
  `setup.cs:14369-14412 [@c26a8a4]`).

### Why per-MAC over upstream-faithful global

mi0bot's effective per-radio semantic comes from per-DB-file scoping
(`database.cs:64 [@c26a8a4]`); multi-radio Thetis users swap files via
`ImportDatabase` (`database.cs:11237 [@c26a8a4]`). NereusSDR achieves the
same property via MAC scoping in one settings file — matches every other
HL2 setting (OcMatrix, HL2 Options, Alex, calibration).

### Tests
22 new unit tests, 275/275 ctest green:
- `tst_hl2_n2adr_persistence` (10) — per-MAC round-trip, multi-MAC
  isolation, 5 migration cases, idempotency.
- `tst_hl2_io_board_tab_n2adr` (7) — `onN2adrToggled` emission, matrix
  mutation, **regression guard for the matrix-wipe-on-Setup-open scenario**,
  no-echo on restore.
- `tst_hl2_step_att_range` (5) — controller signed-range round-trip,
  `setAttenuation` honors negative `min`, `RxApplet` caps→spinbox propagation.

Fixes #105.

## Test plan
- [ ] Connect to HL2; RxApplet S-ATT spinbox scrolls down to `-28 dB` (was 0).
- [ ] Setup → General → Options → Step Attenuator: RX1/RX2 spinboxes accept
      `-28..+32` (was `0..31`).
- [ ] Setup → HL2 I/O → enable N2ADR Filter → quit (Cmd+Q) → relaunch +
      reconnect: checkbox stays ticked, OcMatrix populated.
- [ ] Toggle off → quit → relaunch: comes back unchecked. (Round-trip both
      directions.)
- [ ] Existing HL2 radio with legacy global key gets migrated on first launch
      (verify in `~/.config/NereusSDR/NereusSDR.settings`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -251,6 +251,65 @@ and the chrome layer that surrounds it.
   Verification matrix at
   [`docs/architecture/phase3m-3a-ii-followup-parametriceq-verification/README.md`](docs/architecture/phase3m-3a-ii-followup-parametriceq-verification/README.md).
 
+### Fixed (hermes-filter-debug — HL2 N2ADR + S-ATT bench bugs)
+
+Two HL2 bugs JJ caught while running on a Hermes Lite 2 the day after
+the 3Q chrome layer landed. Both were silent (no warnings, no logs);
+each had a non-obvious root cause that took source-first digging
+through mi0bot to settle.
+
+- **HL2 step-attenuator UI clamped at 0 dB despite the signed
+  `-28..+32` range advertised in `BoardCapabilities`.** The
+  `StepAttenuatorController` clamp at `setAttenuation` was correct,
+  but three UI sites never received the negative bound:
+  `RxApplet::setBoardCapabilities` only updated antenna-button
+  visibility (not the spinbox range); `GeneralOptionsPage` re-ranged
+  the RX1 / RX2 spinboxes from `m_ctrl->maxAttenuation()` but
+  hardcoded `0` for the minimum; `RadioModel::connectToRadio` never
+  pushed the connected board's `attenuator.{minDb, maxDb}` into the
+  controller. Fix wires all three so HL2 sees `-28..+32`
+  (mi0bot `setup.cs:16085-16086 [v2.10.3.13-beta2]`) end-to-end on
+  every connect; legacy `0..31` boards unaffected.
+
+- **N2ADR Filter board enable lost on app restart — and the Setup
+  tab actively destroyed the persisted state.** Write side
+  (`Hl2IoBoardTab::onN2adrToggled`) wrote to global
+  `"hl2IoBoard/n2adrFilter"`; `HardwarePage::restoreSettings` read
+  per-MAC `hardware/<mac>/hl2IoBoard/n2adrFilter`. Per-MAC was always
+  empty, so restore defaulted `checked = false`, then unconditionally
+  fired `onN2adrToggled(false)` — wiping the global key AND the
+  OcMatrix every time the user opened Setup → HL2 I/O. Fix routes
+  the write through `settingChanged()` → `HardwarePage::wire()` →
+  `setHardwareValue(mac, ...)` (the path every other Hardware sub-tab
+  uses). `RadioModel::connectToRadio` reads per-MAC.
+  `restoreSettings` is no-op when the key is absent (a missing key
+  means "no explicit user intent yet" — leave the matrix alone).
+  One-shot migration `AppSettings::migrateLegacyN2adrFilter`
+  (called from `main.cpp`) copies any pre-existing global value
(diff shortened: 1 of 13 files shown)


boydsoftprez:
Thanks @chatgpt-codex-connector — both flags addressed in `5d6f254` (now `4c4fdaa` after a rebase onto main for #159's parametric-EQ landing).

**P1 — Preserve legacy N2ADR key when no HL2 targets migrate** ✅
`migrateLegacyN2adrFilter` now tracks an `hl2Count` separate from `migratedCount` and only deletes the legacy global once at least one HL2 saved radio has been seen during the sweep. Users who enabled N2ADR via the legacy code path before saving any HL2 (or with only a manual-IP entry whose `boardType` is still `Unknown` until first probe) keep the global; a future migration run on the next launch picks it up after the user probes their HL2 and a real `boardType` is recorded.

3 new tests in `tst_hl2_n2adr_persistence`:
- `migrationPreservesGlobalWhenNoHl2Saved` — non-HL2 saved radio only → global retained.
- `migrationPreservesGlobalForUnknownBoardManualEntry` — synthetic `manual-<ip>-<port>` entry with `boardType = Unknown` → global retained.
- `migrationOnNextLaunchPicksUpPreservedGlobal` — first run preserves, user adds HL2, second run completes the migration.

**P2 — Reset N2ADR checkbox state when the key is absent** ✅
`restoreSettings` now sets `m_n2adrFilter` to false on the missing-key path (mirrors the documented default) before returning early. The matrix-non-wipe guard is preserved — `RadioModel::connectToRadio` has already reconciled the matrix to the per-MAC value before this hook runs, so we touch only the checkbox here.

1 new test in `tst_hl2_io_board_tab_n2adr`:
- `restoreSettings_with_missing_key_resets_checkbox_to_false` — pre-set checkbox to checked, restore with empty settings → checkbox unchecked.

Full ctest **283 / 283 green**. New test seam `Hl2IoBoardTab::n2adrCheckboxStateForTest()` / `setN2adrCheckboxStateForTest()` for the P2 verification.

J.J. Boyd ~ KG4VCF

PR #188, opened 2026-05-05
fix(hl2): A-ATT label, N2ADR default-True, dead-checkbox cleanup (#174)

### LINKED ISSUE #174: Auto Step Attenuator on HL2 needs checking
### Bug Description

I noticed when toggling the Auto Step Attenuator on for the HL2, that I can still adjust the step attenuator manually.  For comparison, and a possible improvement, when the Auto Step Attenuator is toggled on, then in the UI, change the step attenuator to show 'A-ATT' instead of staying on the 'S-ATT'.  

https://drive.google.com/file/d/12fHyCvqrkefe42--vR8HYerXTcG3pH5Y/view?usp=drive_link

[nereussdr-20260503-201249.log](https://github.com/user-attachments/files/27328239/nereussdr-20260503-201249.log)

### Steps to Reproduce

1. Open File - Settings
2. Open Options 
3. Toggle on the Enable in the Auto Attenuate RX1 section of the menu dialog 
4. Notice that it doesn't appear to be working with the Step Attenuator displaying 'S-ATT'
5. Also notice the user can adjust the step attenuator despite the fact it is supposed to be in auto attenuate mode  

### Radio Model & Protocol

HL2

### Radio Firmware Version

74

### NereusSDR Version

3.1

### Operating System

Windows 11

### Relevant Logs

```text

```

boydsoftprez:
## Summary
- `S-ATT` to `A-ATT` label flip when auto-attenuate is on, on every radio. Inspired by mi0bot-Thetis `console.cs:21342-21365 [v2.10.3.13-beta2]` (HL2-only there); widened because NereusSDR's `StepAttenuatorController` exposes auto-att on every board.
- N2ADR filter default flipped from `False` to `True` for HL2 boards. Strict port from mi0bot-Thetis `setup.designer.cs:17466-17467 [v2.10.3.13-beta2]` (`chkHERCULES.Checked = true`).
- `OcOutputsHfTab` "N2ADR Filter (HERCULES)" checkbox removed (was a dead writer with no consumer). Replaced with a discoverability hint pointing HL2 users to the real control on Hardware → HL2 I/O Board.

Closes #174.

## Why
- Issue #174 (HL2, firmware 74): toggling Setup → General → Options → "Auto Attenuate RX1 Enable" produced no observable change. The `setAutoAttEnabled` setter emitted no signal, so no label feedback was possible. The reporter's request to swap `S-ATT` for `A-ATT` is in fact mi0bot's exact HL2 convention.
- mi0bot ships with the N2ADR checkbox checked by default. NereusSDR was the outlier shipping it off, generating recurring "N2ADR is broken" reports from users who plugged in the board and didn't know the Setup checkbox needed flipping.
- The `OcOutputsHfTab` "N2ADR Filter (HERCULES)" checkbox was a misleading second surface that wrote to a key nothing read. Users were toggling it and concluding the feature was broken.

## What's wired
- `StepAttenuatorController::autoAttEnabledChanged(bool)` signal added with duplicate-suppression guard.
- `RxApplet::refreshAttLabel` helper considers `(stepOn, autoOn)`. On step-att mode + auto on: `A-ATT`. Step-att mode + auto off: `S-ATT`. Preamp mode: `ATT` regardless of auto state.
- `RadioModel.cpp:1255-1262` and `Hl2IoBoardTab.cpp:1038-1049`: both default to True when the per-MAC `hl2IoBoard/n2adrFilter` key is absent. Existing users who wrote a value keep it.
- `AppSettings::removeOrphanOcN2adrFilter` runs once at startup from `main.cpp` to clean up the dead key.

## Test plan
- [x] Full `ctest` 327/327 green locally on macOS (Apple Silicon).
- [x] New: `tst_rxapplet_a_att_label` (9 cases) verifies the truth table on HL2, Hermes, OrionMKII auto-on, step-off, live toggling, and the new signal contract.
- [x] New: `tst_oc_outputs_hf_no_n2adr_writer` (4 cases) tombstone for the orphan key plus regression guard against a re-introduced dead checkbox.
- [x] Updated: `tst_hl2_io_board_tab_n2adr::restoreSettings_with_missing_key_resets_checkbox_to_default_true` flips the Codex P2 invariant.
- [x] App launches cleanly from a fresh build; `OcOutputsHfTab` shows the new italic hint in place of the dead checkbox.
- [ ] Bench verification on HL2 (firmware 74) and ANAN-G2: confirm `S-ATT` to `A-ATT` flip on auto-attenuate toggle on each board, confirm fresh HL2 default has N2ADR engaged out of the box, confirm legacy users with explicit False keep their setting.

## Risk / silent-flip note
Existing HL2 users who never opened the HL2 I/O Board tab will silently flip from N2ADR off to N2ADR on at next launch. Anyone using HL2 OC pins for non-N2ADR purposes (rare custom external switching) would notice band-correlated pin changes and can toggle the checkbox off. This matches mi0bot's "shipped checked" default.

## Out of scope (filed for follow-up)
- ADC overload badge sensitivity: NereusSDR's 100 ms tick is 4x faster than Thetis's ~400 ms, so the badge fires more eagerly under the same RF conditions. Bench-cross-check against Thetis confirmed this is a NereusSDR-specific divergence. Will be addressed in a follow-up issue with proper bench tuning.
- Auto-att advance algorithm alignment: NereusSDR Classic mode uses severity-scaled bump (`level` 0-5); mi0bot HL2 uses fixed +3 dB. Different feel. Worth aligning but needs bench session.
- OC tab consolidation: N2ADR enable on HL2 I/O Board tab silently mutates the OcMatrix shown on OC Outputs HF/SWL sub-tabs. Three-tab cross-dependency is confusing. Worth a separate consolidation PR.

J.J. Boyd ~ KG4VCF


boydsoftprez (commented):
Automated code review found one issue to address before merging.

src/models/RadioModel.cpp
```diff
@@ -1255,10 +1255,18 @@ void RadioModel::connectToRadio(const RadioInfo& info)
...
+        //   this.chkHERCULES.Checked = true;
+        //   this.chkHERCULES.CheckState = CheckState.Checked;
+        // Users plug in the N2ADR filter board and expect it to "just work"
+        // out of the box; the previous False default forced manual opt-in
+        // and was a recurring support burden.
         const QString n2adrKey = QStringLiteral("hl2IoBoard/n2adrFilter");
         const bool n2adrOn = AppSettings::instance()
                                  .hardwareValue(info.macAddress, n2adrKey,
-                                                QStringLiteral("False"))
+                                                QStringLiteral("True"))
```
boydsoftprez: This default flip is currently applied for every connected radio with a MAC, not just HL2. The block is outside any `info.boardType == HermesLite` / `caps.hasIoBoardHl2` guard, and `applyN2adrPreset()` clears and repopulates the shared OC matrix, so a fresh Hermes/Orion/G2 with no `hl2IoBoard/n2adrFilter` key would get the N2ADR pin pattern persisted on connect. Please scope the `True` default and preset reconcile to HL2, or keep the previous `False` default for non-HL2 boards.

boydsoftprez: Fixed in `b6f3d7d`. Wrapped the n2adrKey read + applyN2adrPreset + matrix.save block in `if (boardCapabilities().hasIoBoardHl2) { ... }`. Confirmed the underlying issue is worse than just the new default: applyN2adrPreset unconditionally wipes the entire OC matrix at N2adrPreset.cpp:73-78 before conditionally repopulating, so the previous False default was already destroying any user-configured non-HL2 OC pin patterns on every connect. The gate fix closes both. Targeted ctest 6/6 still green.

Issue #190, opened 2026-05-05
Step attenuator goes to 32 on HL2


Chris-W4ORS:
### Bug Description

I noticed while adjusting the step attenuator that it goes above 31 to 32 on the HL2 version which opens up the receiver.  

https://drive.google.com/file/d/1ydBDAbpm0qtWLH_QVntY084Z9vRFQIco/view?usp=drive_link

[nereussdr-20260505-072651.log](https://github.com/user-attachments/files/27400066/nereussdr-20260505-072651.log)

### Steps to Reproduce

1. Click on step attenuator's up arrows until it climbs to 31
2. Click on it again and it will go to 32
3. Notice the receiver opens up 

### Radio Model & Protocol

HL2

### Radio Firmware Version

74

### NereusSDR Version

3.1

### Operating System

Windows 11

### Relevant Logs

```text

```

boydsoftprez:
Hi Chris,

PR #194 closes this.  Your bench instinct was right: the spinbox
allowing 32 was an off-by-one upstream bug in mi0bot.

Source check: at mi0bot console.cs:11043 [v2.10.3.13-beta2], the HL2
runtime override widens udRX1StepAttData.Maximum to 32, but the wire
encoding `(31 - userDb) & 0x3F | 0x40` produces wire = -1 at userDb=32
which 6-bit-masks to 0x3F, the same wire value the chip would
interpret as max-LNA-gain wraparound.  The chip cannot physically
distinguish "32 dB ATT" from that wraparound region.  mi0bot's own
InitConsole at console.cs:2111 correctly sets Max=31 (chip range);
the runtime override is the bug.

NereusSDR now caps at +31 (chip-correct) per maintainer approval.
Two related bugs surfaced and got fixed in the same PR while we were
in the area:

1. ATT-on-TX was not pushing to the wire correctly on HL2.  PR #178
   hardcoded the wire byte to 0x40 (max ATT regardless of user
   setting), which forced 31 dB attenuation during every MOX.
   Reverted to the proper user-driven inversion so the radio applies
   exactly the value the user picks.
2. The auto-att system was bumping ATT past the force-31 value during
   MOX (own-TX leakage tripping ADC overflow → bump → 31 → 32).  Now
   gated during MOX so auto-att stays out of the TX-state ATT swap.

Same release timing as #175 once the HL2 hardware bench passes.

J.J. Boyd ~ KG4VCF

PR #194, opened 2026-05-05
fix(hl2): full mi0bot TX parity port + Power page strip-out + ATT-on-TX (#175, #190)

### LINKED ISSUE #175: Adjust the power scale for RF Power and Tune Pwr. for HL2 radios
### Bug Description

The power scale for RF Power and Tune Pwr. for the HL2 radios need to conform to the 0-5W scale instead of 0-100W.  As a suggestion, add a string to the NereusSDR shortcut with -HL2 added so when the application is opened, then the correct scale shows.  As another option, if a HL2 radio is detected, then change the UI to conform to the proper scale. 

[nereussdr-20260503-201249.log](https://github.com/user-attachments/files/27328293/nereussdr-20260503-201249.log)

Image

### Steps to Reproduce

1. Open File - Settings
2. Open the Transmit - Power window
3. Notice the Tune Power (per band) is all set on a 0-100w scale instead of 0-5w scale
4. Exit the menu to return the the main UI
5. Notice in the main UI, the RF Power and Tune Pwr. scales are also on the 0-100w scale 

### Radio Model & Protocol

HL2

### Radio Firmware Version

74

### NereusSDR Version

3.1

### Operating System

Windows 11

### Relevant Logs

```text

```

boydsoftprez:
## Summary

Full mi0bot-Thetis HL2 TX parity port plus several bench-driven fixes that surfaced during testing. Closes #175 (HL2 RF Power and Tune Power scales show 0-100 W instead of 0-5 W) and #190 (Step attenuator goes to 32 on HL2).

After this PR, an HL2 user from mi0bot finds:

- RF Power slider 0-90 step 6 (16-step output attenuator, 0.5 dB/step)
- Tune Power slider 0-99 step 3 (33 sub-steps; 0-51 = DSP audio gain modulation, 52-99 = PA attenuator)
- Setup → Transmit → Power → Tune group ports mi0bot's `grpPATune` exactly (3 drive-source radios + meter combo + Fixed-mode spinbox in dB)
- Per-band Tune Power grid SKU-aware (dB on HL2, watts elsewhere)
- TX volume on the wire matches mi0bot's HL2 sub-step DSP modulation and `(hl2Power * gbb / 100) / 93.75` audio-volume formula
- Tooltip wording rewritten to Thetis upstream `"Transmit Drive - relative value, not watts unless..."` across every SKU
- ATT-on-TX feature works correctly: spinbox follows live applied attenuation during MOX, returns on un-key
- HL2 RX att slider capped at +31 (chip-correct) instead of mi0bot's +32 (off-by-one upstream bug)

Non-HL2 SKUs: tooltip wording change only, no behavior change.

## Source

Per memory rule mi0bot is authoritative for HL2. All ports cite `mi0bot-Thetis [v2.10.3.13-beta2]`. Verified line content unchanged between tag and HEAD `e3375d06`. All `// MI0BOT:` author tags preserved verbatim. Multi-source headers in `TransmitModel.{h,cpp}` per HOW-TO-PORT.md. PROVENANCE rows updated.

### `src/core/HpsdrModel.h`
9 new constexpr per-SKU PA UI constants (slider Max/step, Fixed-mode spinbox bounds, suffix string) + 2 attenuator hardware constants (dB-per-step, step count).

### `src/core/TxChannel.{h,cpp}`
New `setPostGenToneMag(double)` wrapper around WDSP `SetTXAPostGenToneMag`.

### `src/models/TransmitModel.{h,cpp}`
- New `m_txPostGenToneMag` property + signal
- `setPowerUsingTargetDbm` HL2 sub-step DSP modulation in `txMode 1 TUNE_SLIDER` branch
- `computeAudioVolume` HL2 audio-volume formula (placed BEFORE the `gbb >= 99.5` sentinel so HF HL2 bands hit mi0bot's formula instead of the legacy linear path)
- Polymorphic clamp on `setTunePowerForBand` + `setTunePower` + load path (HL2: int 0-99, others: 0-100)
- `setHpsdrModel(HPSDRModel)` setter so TransmitModel can dispatch on connected radio

### `src/gui/applets/TxApplet.{h,cpp}`
- `rescalePowerSlidersForModel(HPSDRModel)` + `updatePowerSliderLabels()` helpers hooked from `RadioModel::currentRadioChanged`
- HL2 dB label formula `(round(drv/6.0)/2) - 7.5` for RF Power, `(slider/3 - 33) / 2` for Tune Power
- Tooltip rewrite across all SKUs

### `src/gui/setup/TransmitSetupPages.{h,cpp}` . Wave 1 strip-out
Source-first IA cleanup. Drops 3 NereusSDR-original UI surfaces with no Thetis upstream:
- Per-band Tune Power 14-spinbox grid (`buildTunePowerGroup`) . Thetis edits via `ptbTune.Scroll` per-band pivot, not a bulk-edit grid
- Reset Tune Power Defaults button (no Thetis equivalent)
- Dead "SWR Protection" slider stub (real SWR controls live in the SWR Protection group below)

Plus drops the duplicate Block-TX checkboxes (live editor at Hardware → Antenna/ALEX → Antenna Control via AlexController per-MAC) and the misplaced AM TX stub on DSP/AM (belongs on Transmit per `grpTXAM`). Adds the `grpPATune` "Tune" group port (3 radios + meter combo + Fixed spinbox). Fixes the TX TUN Meter combo items to match mi0bot verbatim (`Fwd Pwr / Ref Pwr / Fwd SWR / SWR / Off`).

Underlying `m_tunePowerByBand[14]` data structure stays Thetis-faithful (`console.cs:1819, 12094, 17514, 46539 [v2.10.3.13]`); only the bulk-edit grid widget was NereusSDR-spin. Per-band tune editing now happens via the TxApplet Tune slider with band-aware memory, which is the Thetis way.

### `src/core/StepAttenuatorController.{h,cpp}`
- New `m_isMox` field mirrored from `onMoxHardwareFlipped`. Auto-att (Classic + Adaptive) skips during MOX so own-TX leakage tripping ADC overflow doesn't bump `m_attDb` past the intended TX value (e.g. force-31 → 32 → ...)
- `m_savedRxAttDbForTx` field stashes pre-TX RX att so the spinbox tracks the live applied attenuation during TX and restores cleanly on TX→RX
- Emits `attenuationChanged` on the MOX flip so the RxApplet S-ATT spinbox reflects the live state

### `src/core/codec/P1CodecHl2.cpp`
- HL2 user-facing range capped at +31 (was +32). Reverts an off-by-one introduced by mi0bot at `console.cs:11043` where `udRX1StepAttData.Maximum = 32` produces wire = -1 at userDb=32, masking to 0x3F (LNA-gain wraparound region; chip can't physically distinguish "32 dB ATT" there). mi0bot's own `InitConsole` at `console.cs:2111` sets Max=31 (chip-correct); we honor the chip-correct value
- Wire encoding `(31 - userDb) & 0x3F | 0x40` correctly applied for HL2 TX/RX paths (regression from PR #178 hardcoded `0x40` is closed)

### `src/core/BoardCapabilities.cpp`
HL2 + HL2RxOnly attenuator ranges flipped from `[-28, +32]` to `[-28, +31]`. Cite explains the mi0bot upstream bug and the maintainer-approved deviation.

## Test plan

- [x] All 353 ctests green locally on macOS Apple Silicon (was 351 before two new fixes; +2 from the auto-att-during-MOX gate test and the ATT-on-TX spinbox test).
- [x] All 5 attribution verifiers clean (`verify-thetis-headers`, `verify-inline-tag-preservation`, `verify-inline-cites`, `verify-provenance-sync`, `check-new-ports`).
- [x] All 24 commits GPG-signed.
- [ ] **HL2 hardware bench required before ship.** DSP changes drive the actual signal on the wire to the radio. Tests cover the math; the bench confirms the math drives the radio correctly. Coordinator TBD (JJ / Chris W4ORS / Steve).

### Bench checklist (spec §9)

- [ ] HL2 + dummy load + calibrated RF wattmeter or scope into 50Ω.
- [ ] Slider 0-51 sub-step regime: TXPostGenToneMag scales 0.4-0.91, output 0..~0.4 W.
- [ ] Slider 52-99 attenuator regime: full audio gain, output ~1-5 W, no audible glitch at the regime boundary (52-53 on the slider).
- [ ] Slider 99: ~5 W into dummy load.
- [ ] Drive source = Fixed -7.5 dB matches RF Pwr slider at 60.
- [ ] SWR meter reads ~1.0 throughout (no rev-power scaling regression from PR #178).
- [ ] ATT-on-TX ON: S-ATT spinbox jumps from RX value to 31 during MOX, returns on un-key. Stays at 31 across multiple MOX cycles (no climb to 32).
- [ ] ATT-on-TX OFF: S-ATT spinbox stays at user RX value during MOX (no swap).

## Risk and migration

- **HL2 users upgrading**: existing per-band Tune Power values reinterpret as int 0-99 (mi0bot encoding) → dB at the UI boundary. No migration step required (matches mi0bot's polymorphic-key pattern).
- **HL2 users with stored S-ATT value of 32**: silently clamped to 31 on next load. If anyone was operating at exactly +32, they'll be at +31 instead . 1 dB difference in RX sensitivity during TX, no safety impact.
- **PR #178 codec hardcode**: re-introduced the `(31 - userDb)` inversion that PR #178 removed. PR #178's 0x40 hardcode was based on a Thetis-side capture of an active TX state (force-31); using it as the universal default forced max ATT during every MOX regardless of user setting. Post-fix, the wire byte tracks the user's intent.

## Acknowledgments

Thanks to Chris W4ORS for the original report (#175) and the follow-up bug catch (#190) while we were mid-fix.

J.J. Boyd ~ KG4VCF


src/core/StepAttenuatorController.cpp
```diff
@@ -514,7 +541,25 @@ void StepAttenuatorController::onMoxHardwareFlipped(bool isTx)
...
+            //
+            // Use direct emit rather than setAttenuation() to avoid pushing
+            // to connection (connection layer already cleared TX ATT above;
+            // the RX att will get re-pushed on the next CC bank that reads
+            // m_stepAttn — already-stored unchanged from before MOX).
+            //
+            // Then re-sync m_bandState entry in case the band-state RX att
+            // diverged during MOX (defensive — should not happen in 3M-1
+            // single-RX scope, but cheap to verify).
+            if (m_attDb != m_savedRxAttDbForTx) {
```
boydsoftprez: This restore now runs on every non-HPSDR TX->RX transition, but the new stash is only populated in the ATT-on-TX enabled path above. When ATT-on-TX is disabled, the MOX-on branch returns early after sending TX ATT=0, leaving `m_savedRxAttDbForTx` at its default or stale value; then un-keying restores `m_attDb` from that value and emits `attenuationChanged`. A user with S-ATT=15 and ATT-on-TX OFF can key/unkey and see the RX attenuator snap to 0 or a previous TX-window stash, even though the PR bench checklist says ATT-on-TX OFF should leave the S-ATT spinbox at the RX value. Please gate this restore on the path that actually stashed, or seed the stash before the disabled early return; add a test that toggles ATT-on-TX off, sets a nonzero RX ATT, MOX on/off, and asserts `attenuatorDb()` is unchanged.

src/gui/setup/TransmitSetupPages.cpp
```diff
@@ -283,64 +316,177 @@ void PowerPage::buildPowerGroup()
...
+    m_comboTxTunMeter->addItem(QStringLiteral("Fwd SWR"));
+    m_comboTxTunMeter->addItem(QStringLiteral("SWR"));
+    m_comboTxTunMeter->addItem(QStringLiteral("Off"));
+    form->addRow(QStringLiteral("TX TUN Meter:"), m_comboTxTunMeter);
+
+    // ── Fixed-mode tune-power spinbox ──────────────────────────────────────
+    // Bounds / step / decimals / suffix are set in applyHpsdrModel() and
+    // re-applied on every RadioModel::currentRadioChanged.  The default
+    // (FIRST = ANAN100) keeps a sane Watts range until the first connect.
+    m_fixedTunePwrSpin = new QDoubleSpinBox(m_grpPATune);
```
boydsoftprez, line 403: The new Fixed Tune Power spinbox is created and later enabled for `DrivePowerSource::Fixed`, but it is never initialized from `TransmitModel::tunePower()` or connected back to `TransmitModel::setTunePower()`. Selecting "Use Fixed Drive" only changes `tuneDrivePowerSource`; edits to `udTXTunePower` are not persisted or used by the TUNE path, so txMode 1 / Fixed keeps transmitting with the previous/default `m_tunePower`. Please wire valueChanged through the SKU-specific display conversion to `setTunePower()`, mirror `tunePowerChanged` back into the spinbox, and cover that in the PowerPage test.

boydsoftprez (commented):
Manual Codex review found two issues that should be fixed before merging: ATT-on-TX OFF can clobber the RX attenuator value on un-key, and the new Fixed Tune Power control is not wired back to the transmit model.

PR #195, opened 2026-05-06
fix(oc-matrix): persist OC pins per band across restart (#191)

### LINKED ISSUE #191: OC pins per band
### Bug Description

In Hardware config, the OC pins per band are not memorized. We have to reconfigure at  each start. 
Anyway,  very pleased to play with this new software on linux.
Pascal  F6EHP

### Steps to Reproduce

Configure the OC pins per band, quit the software, come back in, and see...no more settings in this windows

### Radio Model & Protocol

HL1 protocol 1 - a home made prototype built in 2015 with the help of Steve KF7O

### Radio Firmware Version

V1.1

### NereusSDR Version

0.3.1

### Operating System

Linux (Ubuntu/Debian)

### Relevant Logs

```text

```

boydsoftprez:
## Summary

- Closes #191 (OC pins per band not memorized across restart on HL1, reported by Pascal F6EHP)
- `OcMatrix::setPin()`, `setPinAction()`, and `resetDefaults()` now persist immediately to `AppSettings` after the in-memory mutation. Previously the OC HF / SWL tab checkbox handlers and the Reset button called these setters with no follow-up `save()`, so user edits never reached `~/.config/NereusSDR/NereusSDR.settings` and were lost on app shutdown.
- Affected every HPSDR board, not just HL1 / HL2.

## Root cause

The only `OcMatrix::save()` call site was the N2ADR connect-time reconcile in [RadioModel::startConnection](src/models/RadioModel.cpp:1280), gated on `hasIoBoardHl2`. User edits anywhere else (HF tab, SWL tab, Reset button) were dropped on quit because `save()` was never called for them.

## Fix locus

- [src/core/OcMatrix.cpp](src/core/OcMatrix.cpp): each setter now calls `save()` after releasing the write lock.
- The existing `m_ocMatrix.save()` after `applyN2adrPreset` in [RadioModel::startConnection](src/models/RadioModel.cpp:1280) is now redundant but harmless; left in as a defensive checkpoint after bulk preset application.

## Test plan

- [x] `tst_oc_matrix` adds 3 regression cases (`setPin_persists_without_explicit_save`, `setPinAction_persists_without_explicit_save`, `resetDefaults_persists_without_explicit_save`); all 3 fail before the fix (red) and pass after (green).
- [x] Full local `ctest`: 353 / 353 passing.
- [ ] Manual verification on real radio (Pascal HL1 or any HPSDR board): toggle a few OC pins per band in Setup → Hardware → OC Outputs → HF, quit, relaunch, confirm checkboxes survive.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/src/core/OcMatrix.cpp
+++ b/src/core/OcMatrix.cpp
@@ -81,6 +81,10 @@ void OcMatrix::setPin(Band band, int pin, bool tx, bool enabled)
         if (slot == enabled) { return; }
         slot = enabled;
     }
+    // Issue #191: persist immediately. The HF / SWL OC tab toggle handlers
+    // call setPin() directly and don't follow up with save(); without this,
+    // user edits never reach AppSettings and are lost on app restart.
+    save();
     emit changed();
 }
 
@@ -116,6 +120,10 @@ void OcMatrix::setPinAction(int pin, TXPinAction action)
         if (m_pinActions[pin] == action) { return; }
         m_pinActions[pin] = action;
     }
+    // Issue #191: persist immediately. The HF tab pin-action lambda calls
+    // setPinAction() directly without an explicit save(); without this,
+    // user edits never reach AppSettings and are lost on app restart.
+    save();
     emit changed();
 }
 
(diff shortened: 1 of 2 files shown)


boydsoftprez (commented):
Manual Codex review found no blocking code issues in this PR. I checked the OcMatrix setter autosave path, MAC scoping/load path, N2ADR preset caller, HF/SWL tab mutation paths, reset flow, and the new regression tests. CI was still pending on this latest head when reviewed.

PR #212, opened 2026-05-07
Phase 3M-4: PureSignal port (HL2 / HermesII / Saturn end-to-end)

boydsoftprez:
## Summary

- Full Phase 3M-4 PureSignal port from design through bench convergence.
  WDSP calcc.c / iqc.c vendored verbatim; PureSignal coordinator,
  PsccPump driver, and per-board DDC routing all source-first against
  Thetis ramdor v2.10.3.13 + mi0bot v2.10.3.13-beta2.
- All P1 PureSignal-capable boards now wired end-to-end: HL2,
  Hermes / ANAN-10 / ANAN-100 (nddc=4), HermesII / ANAN-10E /
  ANAN-100B (nddc=2). HL2 hasPureSignal flipped to true after the
  applyPsDdcConfig consumer landed; pre-PR, HL2 PS was visible-but-broken
  (codec computed cntrl1=4 but no consumer applied it to the wire).
- mi0bot HL2 specifics ported: psFbDdc=2 / txMonDdc=3 PS DDC pair (was
  silently DDC0+DDC1 = wrong stream), HL2 negative-ATT support via
  unified minAttenuation() clamp (NeedToRecalibrate_HL2 + SetNewValues
  branches), HermesII bank 2/3 freq override.
- Bench convergence confirmed on Saturn / ANAN-G2: state=6 (LCALC),
  corrApplied=1, calCount=872, feedbackLevel=164 (inside Thetis target
  window [128, 181]). HL2 bench validation pending; verification matrix
  in docs/architecture/phase3m-4-handoff-bench-debug.md.

## Source-first cite map

| Layer | Upstream cite |
| --- | --- |
| WDSP calcc / iqc | TAPR/Thetis [v2.10.3.13] vendored byte-identical |
| PureSignal coordinator (cmd-state, autoAtt, runcal) | PSForm.cs:235-790 + 1097-1144 [v2.10.3.13] + mi0bot HL2 variants |
| Force-31 SATT predicate | console.cs:29563-29566 [v2.10.3.13] |
| ATT-on-TX master force-enable | PSForm.cs:738 //MW0LGE [v2.10.3.13] |
| HL2 needRecal / clamp lower bound | PSForm.cs:1142-1144 + 786-790 [v2.10.3.13-beta2] |
| P1 bank 2/3 PS freq override (HermesII) | networkproto1.c:484-511 [v2.10.3.13] |
| P1 PsDdcConfig per-board | console.cs:8186-8525 + mi0bot 8408-8488 |
| PS DDC pair (psFbDdc/txMonDdc) per nddc | networkproto1.c MetisRead twist() dispatch |
| P2 CmdHighPriority freq override | network.c:936-945 [v2.10.3.13] |
| HermesII < 103 firmware floor | NetworkIO.cs:138 [v2.10.3.13] (no HL2 floor) |

## Per-SKU PureSignal support matrix

| Board | nddc | hasPureSignal | PS DDC pair | Status |
| --- | --- | --- | --- | --- |
| Atlas | 1 | false | n/a | Correctly disabled |
| HermesLiteRxOnly | n/a | false | n/a | Correctly disabled (no TX) |
| Hermes (orig) / ANAN-10 / ANAN-100 | 4 | true (NEW) | DDC2 + DDC3 | Source-faithful, bench pending |
| HermesLite (HL2) | 4 | true (NEW) | DDC2 + DDC3 | Source-faithful, bench pending |
| HermesII / ANAN-10E / ANAN-100B | 2 | true | DDC0 + DDC1 | Bank 2/3 freq override now correct |
| Angelia / Orion / OrionMKII / Saturn / SaturnMKII / Andromeda | 5 | true | DDC0 + DDC1 (P2 freq override) | Bench-validated on Saturn |

## What's in the diff

- 105 files changed, +20,711 / -678
- 41 commits, all GPG-signed
- ~30 new test files covering PureSignal coordinator, PsccPump, P1
  per-board codec PS branches, P1 bank 2/3 freq override, HL2 ATT
  bounds, P2 codec context updates
- New runtime classes: PureSignal coordinator, PsccPump driver,
  PsFeedbackChannel, TxChannel 22 PS-API wrappers
- New GUI: PsForm modeless dialog (Tools > PureSignal), AmpView,
  PsaIndicatorWidget, two-tone IMD overlay on spectrum, PS-A button
  on TxApplet, HPF Bypass / General Options checkboxes wired
- Vendored WDSP calcc.c + iqc.c + ps_sync_stub.c

## Test plan

- [ ] `cmake --build build -j$(nproc)` clean (no warnings beyond
      pre-existing baseline)
- [ ] `ctest --test-dir build -j8` clean (note: tst_audio_engine_speakers_live_reconfig +
      tst_tx_mic_source are pre-existing -j8 flakes; pass standalone)
- [ ] **Saturn / ANAN-G2 bench:** Tools > PureSignal opens, PS-A on TxApplet
      enables, 2-Tone trigger drives calcc to state=6 (LCALC) with
      feedbackLevel converging into [128, 181] within ~5 s
- [ ] **HL2 bench:** PS-A button visible (was hidden before), PS Setup
      Calibration Information populates non-zero, EP6 frames parse
      cleanly during PS-MOX, SATT auto-attenuate engages and may go
      negative (signed range)
- [ ] **HermesII bench (if available):** bank 2 + bank 3 wire-byte
      capture shows TX freq during PS-MOX (pcap with udp.port == 1024,
      C0=0x04 / 0x06 frequency bytes match TX VFO)
- [ ] **No regressions** on existing P1/P2 test suites (366+ tests in
      PS / codec / step-att / two-tone / HL2 / Hermes families)

Full bench checklist + source-first cite map: `docs/architecture/phase3m-4-handoff-bench-debug.md`

J.J. Boyd ~ KG4VCF


src/core/PureSignal.cpp
```diff
@@ -0,0 +1,1241 @@
...
+    m_hwPeak = m_caps.psDefaultPeak;
+    if (m_tx) {
+        m_tx->setPSHWPeak(m_caps.psDefaultPeak);
+    }
+    // Push the per-board feedback rate to the feedback channel.  Mirrors
+    // Thetis cmaster.cs:535 [v2.10.3.13]:
+    //   puresignal.SetPSFeedbackRate(txch, ps_rate);
+    if (m_tx) {
+        m_tx->setPSFeedbackRate(m_caps.psSampleRate);
+    }
```
boydsoftprez: [P1] Do not pass the HL2 sentinel to WDSP

`psSampleRate=0` is being used as an HL2 sentinel, but this path passes it straight into `SetPSFeedbackRate`. Upstream `cmaster.cs` always calls `SetPSFeedbackRate(txch, ps_rate)` with `ps_rate=192000`; calcc stores that as `a->rate` and rebuilds its delay buffers from it. On HL2 this currently makes calcc run with a zero feedback rate, so timing/delay behavior is broken before calibration even gets useful samples. Resolve the sentinel to a real rate before calling WDSP, and test the HL2 path specifically.

src/models/RadioModel.cpp
```diff
@@ -1870,6 +1887,178 @@ void RadioModel::connectToRadio(const RadioInfo& info)
...
+            // network.c:936-945 [v2.10.3.13] gate is
+            // (ptt_out && puresignal_run)).  Without this, DDC0/DDC1
+            // stay tuned to RX freq during TX and never see the actual
+            // TX signal — the bench shows DDC1's PA-feedback ADC
+            // returning ~zero envelope because nothing's at RX freq.
+            // From Thetis PSForm.cs:246 [v2.10.3.13]:
+            //   NetworkIO.SetPureSignal(1);
+            connect(m_pureSignal.get(), &PureSignal::autoCalEnabledChanged,
+                    m_connection, &RadioConnection::setPuresignalRun,
+                    Qt::UniqueConnection);
```
boydsoftprez: [P1] Port PSEnabled fan-out, not AutoCal only

The original `PSEnabled` setter is the fan-out that calls `UpdateDDCs`, `SetPureSignal`, sends high priority, updates audio routing, loads the router bit, and flips `PSRunCal`; the state machine toggles `PSEnabled` for auto-cal, single-cal, restored corrections, stay-on, and turn-off. Here those side effects are wired only to `autoCalEnabledChanged`, so Single Cal/Restore with PS-A off only set calcc flags and never enable the radio/DDC routing needed by PsccPump. Add a PSEnabled-equivalent signal/helper and drive ReceiverManager/RadioConnection from that source state, not from AutoCal alone.

src/core/PureSignal.cpp
```diff
@@ -0,0 +1,1241 @@
...
+            // to 1 to avoid the divide-by-zero pathological case.
+            const int safeFb = (fbLevel < 1) ? 1 : fbLevel;
+            ddB = 20.0 * std::log10(static_cast<double>(safeFb) / 152.293);
+            if (std::isnan(ddB)) {
+                ddB = 31.1;
+            }
+            if (ddB < -100.0) ddB = -100.0;
+            if (ddB > +100.0) ddB = +100.0;
+        } else {
+            ddB = 31.1;
```
boydsoftprez: [P2] Restore the HL2 auto-attenuation clamps

The mi0bot HL2 branch deliberately clamps abnormal `ddB` to +/-10 dB and uses +10 when feedback is not OK; only non-HL2 boards use 31.1 and +/-100. This shared implementation applies the legacy math to HL2, so a zero/stale feedback reading can drive HL2 straight to its -28 dB floor instead of the source's gentler -10 dB correction. Branch on the HL2 capability here and port the exact mi0bot fallback values.

src/core/PureSignal.cpp
```diff
@@ -0,0 +1,1241 @@
...
+        //   _autoON = false;
+        //   _performing_single_cal = true;
+        //   puresignal.SetPSControl(_txachannel, 1, 1, 0, 0);
+        //   if (!PSEnabled) PSEnabled = true;
+        //   _cmdstate = eCMDState.SingleCalibrate;
+        // PSEnabled=true → calcc.runcal=1 (see Off-case comment above).
+        m_autoON = false;
+        m_tx->setPSControl(/*reset=*/1, /*mancal=*/1,
+                           /*automode=*/0, /*turnon=*/0);
+        m_tx->setPSRunCal(1);
```
boydsoftprez, line 1064: [P2] Re-add single-cal retry loop

Upstream sets `_performing_single_cal=true` when Single Cal starts, then in StayON retries `_singlecalON=true` up to five times when the feedback level is still out of range. This port starts a single calibration and can report completion once corrections apply, but it never records the single-cal attempt or retries bad-feedback runs. That makes the button tooltip/source behavior inaccurate and leaves marginal calibrations unretried.

src/core/PureSignal.cpp
```diff
@@ -0,0 +1,1241 @@
...
+    }
+
+    // Step 3: CorrectionsBeingApplied flag — info[14] == 1 per
+    // PSForm.cs:1100-1102 [v2.10.3.13].  Drives the Save-button
+    // gating + PS-state colour box.
+    const bool newCorrApplied = (newInfo[14] == 1);
+    if (newCorrApplied != m_correctionsApplied.exchange(newCorrApplied)) {
+        // Same correctingChanged signal carries this — UI subscribers
+        // poll correctionsBeingApplied() and isCorrecting() separately.
+        emit correctingChanged(newCorrApplied);
```
boydsoftprez: [P2] Split correction-applied from correcting

The source has two different predicates: `CorrectionsBeingApplied` gates Save and the CO badge, while `Correcting` is only `FeedbackLevel > 90` and decides lime vs yellow after a correction exists. This signal now carries both meanings, so Save can enable from feedback level alone, disable while corrections still exist, and the CO badge never renders the source yellow state. Add a separate `correctionsBeingAppliedChanged` signal and have the UI combine both booleans like PSForm.cs does.

src/core/PureSignal.cpp
```diff
@@ -0,0 +1,1241 @@
...
+
+void PureSignal::setTint(double db)
+{
+    if (db == m_tint) { return; }
+    m_tint = db;
+    // No direct TxChannel setter today — the Thetis comboPSTint handler
+    // (PSForm.cs comboPSTint_SelectedIndexChanged [v2.10.3.13]) drives
+    // ints / spi changes on the engine via puresignal.SetPSIntsAndSpi.
+    // Wire-through is deferred until the calcc test bench at Task 11.
+    emit tintChanged(db);
```
boydsoftprez: [P2] Wire TINT to SetPSIntsAndSpi

The source combo handler calls `SetPSIntsAndSpi` and updates `_ints/_spi` for all three TINT selections: 16/256, 8/512, and 4/1024. This setter currently only stores the displayed dB value and explicitly defers the engine call, while AmpView keeps using the default buffer dimensions. Selecting 1.1 or 2.5 therefore leaves calcc and AmpView on the wrong interval/sample layout.

boydsoftprez (commented):
Source-first PureSignal review against local Thetis + mi0bot sources. I focused on omissions and places where source semantics were flattened during the port.

src/models/RadioModel.cpp
```diff
@@ -965,6 +982,47 @@ RadioModel::RadioModel(QObject* parent)
...
+        if (m_txChannel) {
+            m_txChannel->setTxFixedGain(iqGain);
+        }
+        if (m_connection) {
+            auto* conn = m_connection;
+            QMetaObject::invokeMethod(conn, [conn, wireDrive]() {
+                conn->setTxDrive(wireDrive);
+            });
+        }
+    });
```
boydsoftprez: [P1] Drop duplicate audioVolumeChanged pump

audioVolumeChanged is already connected to RadioModel::pumpAudioVolume above, and this second listener recomputes the opposite SWR topology. The later queued setTxDrive wins, so SWR foldback is removed from the wire byte and put on IQ gain instead of matching Thetis NetworkIO.SetOutputPower / CMSetTXOutputLevel. Delete this listener and rely on pumpAudioVolume.

boydsoftprez (commented):
Source-first follow-up review after the latest PureSignal fixes.

src/core/PureSignal.cpp
```diff
@@ -0,0 +1,1535 @@
...
+void PureSignal::setDefaultPeaks()
+{
+    // From Thetis PSForm.cs:547-550 SetDefaultPeaks [v2.10.3.13]:
+    //   psdefpeak(HardwareSpecific.PSDefaultPeak);
+    // psdefpeak (PSForm.cs:371-381 [v2.10.3.13]) writes the value into
+    // txtPSpeak, which fires PSpeak_TextChanged (PSForm.cs:792-803),
+    // which calls puresignal.SetPSHWPeak.
+    if (m_tx) {
+        m_tx->setPSHWPeak(m_caps.psDefaultPeak);
+    }
```
boydsoftprez: [P2] Finish Default SetPk state update

Thetis SetDefaultPeaks routes through psdefpeak, which updates the text box, fires PSpeak_TextChanged, updates _PShwpeak, pushes SetPSHWPeak, and refreshes the warning. This method only pushes WDSP, leaving m_hwPeak and hwPeakChanged stale; PsForm then mirrors the old value back after pressing Default. Route through setHwPeak(m_caps.psDefaultPeak) and refresh the SetPk warning state.

src/gui/PsForm.cpp
```diff
@@ -0,0 +1,1190 @@
...
+    auto* lblSetPk = new QLabel(tr("SetPk"), grp);
+    g->addWidget(lblSetPk, 2, 4);
+    // PR #212 follow-up bench fix (J.J. KG4VCF, 2026-05-07):
+    // SetPk converted from QLabel (read-only mirror) to QLineEdit (user-
+    // editable) — matches Thetis PSForm.designer.cs:573-582 txtPSpeak which
+    // is a TextBox.  Wired to PureSignal::setHwPeak via editingFinished so
+    // bench-tuned values stick (HL2 + N2ADR users typically need ~0.117 vs
+    // the spec 0.233 default — the AutoAtt loop can't converge on hardware
+    // where the feedback signal differs from spec without this).
+    // From Thetis PSForm.cs:815-827 PSpeak_TextChanged [v2.10.3.13]:
```
boydsoftprez: [P2] Fix CI-blocking PSpeak source cite

Linux CI is failing verify-inline-tag-preservation on this cite and the same PSpeak cite near line 899. Baseline Thetis PSForm.cs:815 is chkPSAutoAttenuate with a MW0LGE tag; PSpeak_TextChanged is baseline line 787, while mi0bot has it at line 815 and UpdateWarningSetPk at 830 with the MW0LGE tag. Correct the source line/source label and preserve the warning tag where applicable.

src/core/PureSignal.cpp
boydsoftprez: Fixed in `28b9457`. `applyBoardCapabilities` resolves the HL2 `psSampleRate==0` sentinel via the `kThetisPsRate=192000` fallback before `SetPSFeedbackRate` (PureSignal.cpp:677-680).

J.J. Boyd ~ KG4VCF

src/models/RadioModel.cpp
boydsoftprez: Fixed in `e5114f2`. `PureSignal::psEnabledChanged` now drives both `ReceiverManager::setPureSignalEnabled` and `RadioConnection::setPuresignalRun` (RadioModel.cpp:2008-2027), so Single Cal / Restore / Stay-on / Turn-off all hit the radio-side fan-out.

J.J. Boyd ~ KG4VCF

src/core/PureSignal.cpp
boydsoftprez: Fixed in `474a321`. AutoAtt branches on `isHl2 = (HermesLite || HermesLite2)` and applies the mi0bot ±10 dB clamp + `+10 on !IsFeedbackLevelOKRange` rule on HL2; non-HL2 keeps the legacy 31.1 / ±100 path (PureSignal.cpp:1405-1428).

J.J. Boyd ~ KG4VCF

src/core/PureSignal.cpp
boydsoftprez: Fixed in `c18e85d`. `_performing_single_cal` + `_performing_single_cal_retries` ported into the `StayOn` branch with the 5-attempt `IsFeedbackLevelOKRange` retry gate (PSForm.cs:688-699 [v2.10.3.13]).

J.J. Boyd ~ KG4VCF

src/core/PureSignal.cpp
boydsoftprez: Fixed in `b7fafaa`. `correctionsBeingAppliedChanged` and `correctingChanged` are distinct signals on distinct predicates: `info[14]` for Save / CO badge, `FeedbackLevel > 90` for the colored state box.

J.J. Boyd ~ KG4VCF

src/core/PureSignal.cpp
boydsoftprez: Fixed in `75b6a35`. `setTint(double)` routes through `setTintIndex(int)` which dispatches `SetPSIntsAndSpi(16,256) / (8,512) / (4,1024)` per case (PureSignal.cpp:561-590), matching PSForm.cs:857-885 [v2.10.3.13].

J.J. Boyd ~ KG4VCF

src/models/RadioModel.cpp
boydsoftprez: Fixed in `fc28cdf`. Removed the duplicate listener at RadioModel.cpp:986-1024 so the canonical `pumpAudioVolume` connection at line 965 is the only audio-volume sink. Wire byte gets the `swr_protect` factor per Thetis NetworkIO.SetOutputPower (NetworkIO.cs:201-211 [v2.10.3.13]); IQ scalar stays clean. `swrFoldback_appliesToIqNotWireByte` regression test passes.

J.J. Boyd ~ KG4VCF

src/core/PureSignal.cpp
boydsoftprez: Fixed in `fc28cdf`. `setDefaultPeaks` routes through `setHwPeak(m_caps.psDefaultPeak)` so `m_hwPeak` cache + `hwPeakChanged` signal flip alongside the WDSP push, mirroring Thetis psdefpeak's chained flow (PSForm.cs:371-381 [v2.10.3.13]).

J.J. Boyd ~ KG4VCF

src/gui/PsForm.cpp
boydsoftprez: Fixed in `67298ff`. Tightened all three cites to `PSForm.cs:787-794 PSpeak_TextChanged [v2.10.3.13]` (PsForm.cpp:466 / 633 / 899). The +/-5 source-window now ends at line 799, excluding the unrelated `//[2.10.3.7]MW0LGE` tag at PSForm.cs:802 on UpdateWarningSetPk's `pbWarningSetPk.Visible`. Tag-preservation passes locally; CI re-running.

J.J. Boyd ~ KG4VCF

PR #219, opened 2026-05-08
Thetis display + DSP parity epic

boydsoftprez:
## What's in this PR

This branch is the **Thetis display + DSP parity epic** — 81 commits that together close the gap between NereusSDR and Thetis on everything you see in the spectrum panel, the DSP options dialog, the meters, and the noise-floor overlays. The work was originally scoped as a v0.3.0 audit (`docs/architecture/thetis-display-dsp-parity-design.md`, `docs/architecture/thetis-display-dsp-parity-plan.md`) and grew over multiple bench sessions.

If you only have time to read one section, read **Noise floor overlay**.

## Channel rebuild infrastructure

The whole epic stands on a new `RxChannel::rebuild()` / `TxChannel::rebuild()` API that lets us swap WDSP channel state without tearing down the audio pipeline. Adds `ChannelConfig`, `RxChannelState`, and `TxChannelState` structs that capture/apply parameters in one round-trip, plus three `RadioModel` coordinators:

- `setSampleRateLive(rate)` — 8-step orchestration to retune sample rate without disconnect
- `setActiveRxCountLive(count)` — toggle RX2 without rebuild
- `dspChangeMeasured(qint64 ms)` — signal for "time to last change" readout in DSP-Options

In-place filter resize / filter-type setters replace heavy rebuilds when only the filter shape changes (`onModeChanged` no longer triggers a full WDSP rebuild). TX live-apply on band/mode change race resolved.

## WDSP analyzer port

The spectrum trace pipeline used to be a NereusSDR-original detector + percentile estimator. This PR ports the actual WDSP `detector()` and `avenger()` from `analyzer.c` verbatim — same bin-to-pixel reduction, same frame averaging, same linear-power side-channel, same window equivalent-noise-bandwidth handling. Single Thetis-faithful pipeline replaces the legacy parallel paths.

## FFT slider + window functions

Setup → Display → Spectrum Defaults → FFT group rebuilt as a Thetis-faithful 3-column grid: Size header + Min/Max endpoints + bin-width + FFT-size readouts on one row, Window combo on row 2, Hz/bin Target on row 3. Seven window types (Rectangular, Blackman-Harris 4T, Hann, Flat-Top, Hamming, Kaiser, Blackman-Harris 7T). Window changes apply immediately. Overlap support added so FFT frame rate decouples from FFT size.

K-based auto-zoom replans the FFT inversely with bandwidth (`targetSize = baseline × sampleRate / bwHz`), capped at 65536 to bound the buffer-fill pause, hysteresis [0.66, 1.5] to avoid replan thrash. Avengers cleared on FFT size change to prevent cross-resolution ghosting in the waterfall, and a new **8-frame crossfade** dissolves the trace into the new resolution instead of snapping ("trace drops in from sky" complaint).

New **Hz/bin Target** spinbox overrides the bins-in-window default with `targetSize = sampleRate / hzPerBin` — zoom-independent FFT sizing, matches Thetis behaviour exactly when on (Thetis only ever set FFT size from the slider). Default OFF preserves NereusSDR's auto-zoom enhancement.

## Noise floor overlay

Source-first port of Thetis `processNoiseFloor` (display.cs:5866-5912). Replaces the percentile-of-rendered-pixels estimate (which floated above the visible noise on signal-dense bands) with the actual quiet-bin-mean + lerp algorithm:

- Per-pixel accumulator: bins below the previous-frame estimate get linear-power averaged
- `m_nfFftBinAverage` — current per-frame estimate (clamped, drifts up when not enough quiet samples)
- `m_nfLerpAverage` — exponentially smoothed across `framesInAttack` frames

Renders the line + 8×8 box + dBm text + vertical connector exactly per Thetis layout (display.cs:5219-5448). Default Hz/bin sensitivity = 3 (matches Thetis); earlier draft used 50 which made the formula unreachable and the trace perpetually drifted to +200 dBm.

Fast-attack triggers wired from band change, freq jump > 500 kHz, and MOX transitions per display.cs:879-905. Auto-clear gated on `|fft - lerp| < 1.0` AND elapsed > 1000 ms per display.cs:5904-5908.

Setup → Display → Spectrum Defaults gains:
- NF Shift spinbox (`-12..+12 dB`, Thetis `_fNFshiftDBM`)
- Line / Text / Fast-attack colour pickers (defaults retuned for visibility on cyan trace: magenta / yellow / light grey)
- Line width spinbox (`1..5 px`)
- All persisted in `#AARRGGBB` format aligned with the rest of the codebase (zero-alpha values get migrated to defaults — fixes a corruption from an earlier `#RRGGBBAA` mismatch)

Structural fix to `dbmToY` / `dbmToYf`: when the band plan strip is enabled, both helpers reserve `m_bandPlanFontSize + 4` pixels at the bottom of the rect so the dBm floor maps to the **top** of the band plan strip, not the panel bottom. Every spectrum-coordinate consumer (trace, NF line/box/text, peak hold, peak blobs, dBm scale labels, grid horizontal lines) inherits the carve-out — they all derive Y through these helpers.

## Spectrum overlays

- Active Peak Hold trace with per-bin decay + distinct trace colour (`ActivePeakHoldTrace` class)
- Peak Blobs class with hold + drop semantics (`PeakBlobDetector`); per-frame `ResetBlobMaximums` port for hard cut + hold-off reset
- Cursor frequency unified — the Setup checkbox and the SpectrumOverlayPanel button now drive the same visibility flag (`m_showCursorFreq`); format always MHz (`14.2700 MHz`)
- Normalized-trace toggle wired (was a TODO) — applies `-10·log10(binWidthHz)` shift inside `dbmToY` so the entire spectrum plus every overlay renormalises to 1 Hz reference bandwidth in lockstep
- Peak Value + Bin Width overlays added to the GPU path (were CPU-only — silently did nothing in default Metal renderer)
- FPS-spin readout sticking at 30 after dialog reopen — fixed via defensive mirror in `pushFps` (root cause was `loadFromRenderer`'s `QSignalBlocker` on the slider blocked the bidirectional sync from `makeSliderRow`)

## DSP-Options page

New `Setup → DSP → Options` page with 18 controls in Thetis-faithful 3-column / 4-row layout. Per-mode buffer / filter / filter-type live-apply via the Phase 1 rebuild API. Filter Impulse Cache toggles (load/save to `~/.config/NereusSDR/impulse_cache.bin`). High-resolution filter characteristics with `FilterDisplayItem` high-res render path. Warning icon validity logic ported from `console.cs:38797-38807`. "Time to last change" readout subscribes to `dspChangeMeasured`. RX/TX combo split (settings schema v5 separates Buffer/Filter Size into Rx/Tx).

## Multimeter + meter improvements

- New `Display → Multimeter` page with full S/dBm/µV unit-mode fan-out across all `MeterItem` subclasses
- Configurable `MeterPoller` interval + averaging
- `HistoryGraphItem` duration setting + signal history wire-up
- `Appearance → MeterStylesPage` gains "Small mode/filter on VFOs" toggle
- `Diagnostics` page renamed "Logging & Performance" + 3 perf toggles
- `Hardware → RadioInfoTab` gains ANAN-8000DLE volts/amps (cap-gated)
- `General → Options` gains CPU meter rate spinbox

## Grid & Scales

- NF-aware grid: `Adjust grid min to track noise floor` toggle + offset spinbox, auto-tracks the live NF estimate
- Per-band NF-estimate priming (NereusSDR-original) — eliminates cold-start visual jump on band change
- Active-band indicator promoted to a centered section header (`◆ Editing per-band grid: 20m ◆`) above an HRule divider; band name folded into row labels (`dB Max (20m): / dB Min (20m):`) so the association is visible from any focal point
- Dialog round-trip fixed — the toggle and its dependents (offset, maintain delta) now restore state on every Setup reopen. Two compounding bugs: `loadFromRenderer` early-returned on null `pan` even though NF tracking lives on `SpectrumWidget` alone, and `setChecked` inside a `QSignalBlocker` skipped the toggle handler that gates the dependents' enabled state.

## Settings migration

`SettingsSchemaVersion=3` retires four obsolete display keys. `SettingsSchemaVersion=5` splits DSP-Options Buffer/Filter Size into Rx/Tx. CHANGELOG v0.3.0 entry. PROVENANCE audit added 22 new rows. 7 new verification matrix files at `docs/architecture/phase3g8-verification/` (~79 rows total). Final cleanup audit caught 13 missing `//MW0LGE` author tags in NF-aware grid code.

## Misc fixes

- `transmit-model`: pre-connect Mic_Source survives app restart
- `audio`: VAX TX → TxChannel mic-source wiring
- `hal-plugin`: self-recovering shm mapping + diagnostic os_log

## Test plan

- [ ] **NF overlay** — line + 8×8 box + dBm text appear at visible noise floor, settles ~30 s on cold start, instant on band/freq/MOX (gray during fast-attack)
- [ ] **Replan crossfade** — zoom across hysteresis stops, trace dissolves into new resolution instead of snapping
- [ ] **Hz/bin override** — set 5 Hz/bin, zoom in/out, Hz/bin readout stays constant; OFF preserves bins-in-window default
- [ ] **Cursor frequency** — Setup checkbox + overlay-panel button drive same flag, format always `14.2700 MHz`
- [ ] **Normalized trace** — toggle shifts entire spectrum + every overlay
- [ ] **Peak value + bin width overlays** render in default GPU mode
- [ ] **Grid & Scales band indicator** updates live; row labels include band name
- [ ] **Band plan boundary** — NF line, trace bottom, peaks, grid all stop at top edge of band plan strip
- [ ] **NF tracking persistence** — enable, set offset, close Setup, reopen; toggle ON, dependents enabled, offset restored
- [ ] **Active Peak Hold + Peak Blobs** — visible in GPU path; hold/drop/decay per Thetis semantics
- [ ] **DSP-Options live-apply** — change buffer / filter size, time-to-last-change updates
- [ ] **Multimeter unit-mode fan-out** — S → dBm → µV updates all bound items
- [ ] **Sample-rate live-apply** — 48k → 96k → 192k → 384k while listening
- [ ] **Active-RX-count live-apply** — RX2 toggle without disconnect
- [ ] **Connect to ANAN-G2 / HL2** — no regressions in TX, PA setup, multimeter, CPU meter

J.J. Boyd ~ KG4VCF


src/models/RadioModel.cpp
```diff
@@ -5672,6 +5732,364 @@ QString RadioModel::connectionSampleRateText() const
...
+        m_wdspEngine->rebuildRxChannel(0, rxCfg);
+    }
+    if (m_txChannel) {
+        ChannelConfig txCfg;
+        // TX channel uses fixed DSP rate (96 kHz) and output rate from the
+        // connection (P1 = 48 kHz, P2 = 192 kHz).  Only the input rate tracks
+        // the wire sample rate.
+        txCfg.sampleRate = newRateHz;
+        txCfg.bufferSize = 256;  // From WdspEngine::createTxChannel default
+        m_wdspEngine->rebuildTxChannel(1, txCfg);
```
boydsoftprez, line 5813: **[P1] Refresh TX channel after rebuild**

`rebuildTxChannel` destroys and replaces the `TxChannel` unique_ptr, but `RadioModel` keeps using the old raw `m_txChannel`. The next block calls `moveToThread` on that stale pointer, so a live sample-rate change with TX initialized can hit a use-after-free. Refresh `m_txChannel` from `WdspEngine` after rebuild, or have `rebuildTxChannel` return the new pointer and re-run the TX wiring.

src/models/RadioModel.cpp
```diff
@@ -5672,6 +5732,364 @@ QString RadioModel::connectionSampleRateText() const
...
+        // from AppSettings in the production DspOptionsPage path; here we use
+        // the WDSP defaults so the caller only has to pass the rate).
+        m_wdspEngine->rebuildRxChannel(0, rxCfg);
+    }
+    if (m_txChannel) {
+        ChannelConfig txCfg;
+        // TX channel uses fixed DSP rate (96 kHz) and output rate from the
+        // connection (P1 = 48 kHz, P2 = 192 kHz).  Only the input rate tracks
+        // the wire sample rate.
+        txCfg.sampleRate = newRateHz;
```
boydsoftprez, line 5811: **[P1] Preserve TX input/output rate geometry**

The live rebuild path sets `cfg.sampleRate` to the new RX wire rate and `rebuildTxChannel` uses that same value for both TX input and output. Initial TX setup uses 48 kHz mic input, 96 kHz DSP, and `connection->txSampleRate` output, which is 192 kHz on P2. Rebuilding with one rate can break TX audio timing/buffer sizing after sample-rate live apply.

src/core/WdspEngine.cpp
```diff
@@ -717,4 +868,102 @@ TxChannel* WdspEngine::txChannel(int channelId) const
...
+
+    qCInfo(lcDsp) << "Rebuild: opened TX channel" << channelId
+                  << "bufSize=" << cfg.bufferSize
+                  << "rate=" << cfg.sampleRate;
+#endif
+
+    // Construct a new TxChannel C++ wrapper.
+    // outputBufferSize = inputBufferSize × outputSampleRate / inputSampleRate
+    // (same ratio math as createTxChannel — integer multiply-then-divide is safe).
+    const int outputBufferSize = cfg.bufferSize * cfg.sampleRate / cfg.sampleRate;
```
boydsoftprez, line 954: **[P1] Mirror createTxChannel DEXP and parent setup**

`rebuildTxChannel` closes and erases the old TX wrapper, but unlike `createTxChannel` it does not destroy/recreate DEXP, does not recreate or attach the DEXP buffer, and constructs `TxChannel` with `this` as a Qt parent. That parent contradicts the create path's thread-affinity invariant and will prevent a later `moveToThread`; the missing DEXP setup also leaves VOX/DEXP state stale or absent after rebuild.

boydsoftprez (commented):
Two additional findings could not be anchored as inline comments because the referenced lines are not part of the PR diff:

**[P2] Radio Info changes never call live-apply** (`src/gui/setup/hardware/RadioInfoTab.cpp:337`)

The PR adds `setSampleRateLive` and `setActiveRxCountLive`, but the Setup Hardware Radio Info controls only emit `settingChanged`, which `HardwarePage` persists. I do not see a call path from these UI changes into either live-apply method, and the banner still tells users to reconnect, so the advertised live apply is unreachable from this UI.

**[P2] RX DSP worker still routes every receiver through RX0** (`src/models/RxDspWorker.cpp:109`)

`processIqBatch` ignores `receiverIndex`, always fetches `rxChannel(0)`, and always sends audio to slice 0. Once `activeRxCountLive` creates and activates additional RX channels, any receiver-1+ IQ delivered here will be accumulated and processed as RX0 instead of its own channel, corrupting multi-RX audio rather than enabling the new count.

PR #238, opened 2026-05-12
Phase 3J-2 (spot system) + Phase 3R (RADE peer mode)

boydsoftprez:
## Summary

The full delivery of two phases plus their on-air bench-fix tail. Around 70 commits since v0.4.0.

**Phase 3J-2** is a standalone client-side spot pipeline — DX cluster, RBN, WSJT-X UDP, SpotCollector, POTA, FreeDV Reporter, PSK Reporter — feeding a panadapter overlay, a Spot List, and DXCC-aware coloring. Not gated on TCI.

**Phase 3R** is RADE wired as a first-class DSP mode alongside SSB / AM / FM / CW, not faked through DIGU or a virtual audio bus.

Version + tag are intentionally not bumped here. The `/release` skill picks those at release time.

### Spot system (Phase 3J-2)

**Seven sources, one place.** New Tools → Spot Hub dialog (Ctrl+Shift+S) hosts every ingest:

- DX Cluster — telnet to dxc.k1ttt.net, ng7m-1, or your favorite node
- RBN — telnet.reversebeacon.net with per-spot SNR
- WSJT-X — UDP listener on 2237 (the default WSJT-X port)
- SpotCollector — DXLab's UDP push protocol
- POTA — HTTPS poll against api.pota.app every 60 seconds
- FreeDV Reporter — Socket.IO connection to qso.freedv.org
- PSK Reporter — IPFIX UDP, send-only, 5-minute timer

Each source has its own tab with its own Connect/Start button, status label, console stream, and auto-connect toggle. Auto-connect state persists per source and restores at app launch.

**One Settings tab** at the front of the dialog sets your callsign + grid + status message once, then propagates to every source.

**A unified Spot List** aggregates all seven sources into a single sortable table — band filter pills (160m through 2m), source filter pills, click any row to tune.

**Spots on the panadapter.** Callsign labels appear on the spectrum overlay at each spot's frequency, with collision-avoidance multi-level stacking. When too many spots stack on one frequency you get a "+N" cluster badge — click it to see the overflow in a menu.

Mouse over a label → hand cursor + tooltip with full metadata (call, freq, mode, source, spotter, comment, timestamp). Right-click → Tune to / Copy Callsign / Lookup on QRZ / Remove Spot. Hover a row in the Spot List → matching label on the panadapter lights up with a yellow halo; works both directions.

**Per-source visibility toggles** on the Display tab let you turn off RBN labels (or any other source) on the panadapter without affecting the Spot List.

**DXCC color provider** integrates cty.dat (refreshed quarterly in-tree) and an ADIF log you provide. Spots from countries you've never worked on this band show in one color, countries needed for an award show in another, plain-ole-worked show in another. Standard 4-tier resolver matching how AetherSDR does it.

**Dedicated FreeDV Reporter dialog** (Tools → FreeDV Reporter, Ctrl+Shift+R) — 14-column live station map separate from the Spot List. TX stations highlight red, RX-decoding stations highlight green. QSY support: right-click another station to ask them to move freq. MRU status messages with Save / Send / Clear. Distance + heading auto-computed from your grid. Hide-from-view toggle for SWLs. Frequency display in MHz or kHz. 2-hour idle auto-removal. Sort column / column widths persist across restarts.

**PSK Reporter** is send-only by design (matches freedv-gui). Click Start → 5-minute auto-send timer arms. RADE EOO decodes and WSJT-X decodes both flow in. Spots appear on pskreporter.info filtered to your callsign within ~10 minutes.

### RADE peer mode (Phase 3R)

RADE is the new RX/TX mode in Mode menu — RADE-U and RADE-L for upper and lower sideband variants, mirroring the USB/LSB convention. Same dial frequency, 1700 Hz modem bandwidth, full neural-codec voice quality.

**Vendored librade + Opus** at `third_party/rade/` (BSD-2-Clause, ~9 MB embedded neural weights). No external model file ships, no post-install download step.

**RX is live.** Tune to a RADE frequency, switch to RADE-U (or -L), and decoded speech reaches the speakers. The VFO flag grows a new SNR row showing the live SNR plus a sync indicator. When the modem extracts the speaker's callsign from the End-of-Over text channel, the SNR row updates to show `KK7GWY ● 12dB` instead of `RADE ● 12dB`. The callsign sticks while you stay in RADE; clears when you flip sidebands or leave RADE; survives brief sync fades (debounced 2 seconds).

**TX is live too.** Mic audio routes through the RADE encoder, the 24 kHz stereo modem output goes out the radio, and TUNE in RADE mode works (routes through the WDSP modulator). Tested on-air on ANAN-G2 with remote decoders confirming the signal.

**RadeApplet** docks in the right column when RADE is active: profile combo, sync indicator, Reset Vocoder button. **A new RADE factory profile** in the mic processing chain (Leveler enabled; ALC / CFC / CESSB / Phase Rotator all bypassed — matches what RADE expects).

**`RxDecodeModel`** now sources from both WSJT-X UDP decodes and RADE callsign-over-EOO decodes — the "what my radio just heard" panel covers both.

### Bench-fix tail

The above two phases shipped to bench testing as drafts. These are the gaps that surfaced and got closed before release:

- **First MOX produces RF without TUNE-priming first.** Audio-volume scalar was zero at app launch, only set by power-slider movement or TUNE. Now seeded at radio-connect time.
- **Spot Hub Connect/Start buttons actually do things.** Ten per-tab lifecycle signals were emitting into the void — auto-connect bypassed them so it looked half-functional. All ten wired.
- **DX cluster spots actually show.** Parser only knew the AetherSDR-ported format, not the DXSpider format the most common cluster hosts use. Added a fallback regex.
- **Spots on the panadapter.** The bridge from data model to renderer was missing entirely (the AetherSDR `refreshSpots` lambda never carried over). Added.
- **Spot List populates from launch on auto-connect.** The list's table model used to live inside the dialog and only existed once you opened the dialog; spots from auto-connect were lost. Moved up to live with the rest of the spot infrastructure.
- **Status labels + buttons reflect real state on every spot tab.** Was only wired on the FreeDV tab; now everywhere.
- **FreeDV Reporter distance/heading columns populate.** The model's "our grid" was never being told what your grid is. Now seeded from Settings.
- **FreeDV Reporter status messages reach the dashboard reliably.** The qso.freedv.org server sometimes skips the handshake event we were waiting for. Push at a more reliable point.
- **VFO spin doesn't DoS qso.freedv.org.** New 7-second dwell on frequency publishes, with a 100 kHz band-jump fast-path so band changes don't lag.
- **TX/RX row highlights win over selection.** Qt's selection brush was clobbering the colored row backgrounds. Fixed.
- **PSK Reporter Start button arms the timer.** Was wired to nothing.
- **Spot overlay has hover + right-click + halo + per-source visibility.** Adversarial audit against AetherSDR found ten missing interaction features; closed all ten.
- **First-time setup banner** on the Settings tab when no callsign is configured. Placeholder text rewritten from passive ("your callsign") to action ("Enter Callsign Here").

## What stays out of scope

- HL2 RADE bench verification — gated on closure of the HL2 ATT/filter safety audit. Tracked separately.
- RADE-on-A while SSB-on-B multi-slice — exploratory; waits on Phase 3F multi-panadapter.

## Build + packaging

`third_party/rade/` and `third_party/r8brain/` added to the vendored dependency set. CMake glue builds RADE + Opus + LPCNet via ExternalProject. Cold CI build adds ~90 s on first job; incremental builds cached. AppImage / DMG / Windows installer artifacts all carry the embedded neural weights.

## For existing users

No action required. Saved radios, mic profiles, DSP settings, container layout, spectrum / waterfall settings all carry forward. The new RADE factory profile (Profile #22) appears automatically alongside the existing 21. Spot system keys default to inactive on first launch — nothing connects to anything until you enable it in Spot Hub.

## Test plan

- [ ] Cold launch with auto-connect on DX Cluster + RBN → Spot List populates immediately
- [ ] POTA → Start → status reads "Polling api.pota.app", `pota.log` records activations
- [ ] PSK Reporter → enter callsign+grid → Start → make decodes → within 10 min, your spots appear at pskreporter.info
- [ ] Right-click a panadapter spot label → context menu with Tune / Copy / QRZ / Remove
- [ ] Hover a spot label → hand cursor + tooltip with metadata
- [ ] Hover a Spot List row → matching panadapter label gets yellow halo
- [ ] Display tab → uncheck RBN → RBN labels vanish from panadapter, Spot List unaffected
- [ ] FreeDV Reporter dialog → Distance + Hdg columns non-zero once a grid is configured
- [ ] RADE mode, listen to an EOO decode → flag switches to `<callsign> ● <snr>`
- [ ] RADE TX → mic audio reaches remote receivers, confirmed by external decoder
- [ ] Spin VFO inside one band → single FreeDV freq_change packet 7s after stop. Jump bands → packet immediate
- [ ] Cold launch with no callsign → orange first-run banner on Settings tab
- [ ] First MOX after launch (no prior TUNE) → modulation immediate

J.J. Boyd ~ KG4VCF


boydsoftprez:
## Review notes

I did a broad pass over the changed-file surface, then went deep on the highest-risk areas: CI/build integration, RADE/WDSP mode switching, DSP worker flow, RADE channel lifetime, and the spot-system model/client paths. I did **not** line-review every added generated/vendor RADE data line; this PR is very large and GitHub's diff view also exceeded its line limit.

### Findings

- **P0: RADE Opus build is not portable in CI.** `third_party/rade/cmake/BuildOpus.cmake` hard-codes `./autogen.sh && ./configure ...` inside `ExternalProject_Add`. The macOS job does not install the autotools stack needed by `autogen.sh`, so it fails with `autoreconf: command not found`. On Windows, the ExternalProject step runs under `cmd.exe`, so `./autogen.sh` fails as `'.' is not recognized...`. This is currently blocking macOS/Windows/CodeQL-style builds. A fix probably needs either a CMake-native/prebuilt Opus path, or explicit CI/toolchain setup plus an explicit shell invocation for the configure step on Windows.

- **P1: Runtime switching into RADE appears to leave RX silent.** `SliceModel::setDemodMode()` destroys the WDSP `RxChannel` when switching from a WDSP mode into RADE, but `RxDspWorker` now requires that `RxChannel` to exist before it processes or forks audio into RADE. That means USB/LSB -> RADE can drop every incoming chunk before `RadeChannel::processIq` is reached. With the new “WDSP demod front-end” flow, the mode switch likely needs to keep the WDSP RxChannel alive and change only routing/gating behavior.

### Areas worth going deeper next

- **RADE channel lifetime/threading.** `RxDspWorker` queues work against a raw `RadeChannel*` while channels can be destroyed from the model/engine path. I saw enough to be suspicious, but not enough for a firm finding. This deserves a focused pass for use-after-free races and whether `QPointer`, queued teardown, or DSP-thread serialization is needed.

- **Actual RADE mode transition matrix.** Test cold start in RADE, USB/LSB -> RADE, RADE -> USB/LSB, rapid toggles, and persisted session restore. The code has different paths for startup and runtime transitions, so one path can pass while another silently drops audio.

- **Spot-system UI refresh semantics.** I skimmed PSK Reporter/FreeDV station ingestion and distance/heading recomputation. It would be worth verifying whether changes to local grid/identity immediately repaint existing station rows and spot metadata, especially if the dialog is already open.

- **CI coverage shape.** Local app build passed for me with `NEREUS_GPU_SPECTRUM=OFF`, but this configure path does not build tests by default. Given the size of the spot/RADE integration, I’d consider adding at least one CI job/configuration that enables the new tests and exercises the non-GPU app build.

Local verification performed:

```text
cmake -B build-review -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DNEREUS_GPU_SPECTRUM=OFF
cmake --build build-review -j$(sysctl -n hw.ncpu)
```

Both local commands completed successfully on macOS.

boydsoftprez:
## Deeper adversarial pass

I went deeper on the RADE transition matrix, WDSP/RADE threading boundary, spot identity propagation, and the real CI/test shape. These are additive to my earlier comment.

### Additional findings

- **P1: persisted/kept RADE modes can be passed directly into WDSP `SetRXAMode`.** `DSPMode::RADE_U`/`RADE_L` are explicitly documented as non-WDSP modes (`src/core/WdspTypes.h:159-186`), but the startup path now always creates a WDSP `RxChannel` and immediately applies `m_activeSlice->dspMode()` to it (`src/models/RadioModel.cpp:2655-2661`). `RxChannel::setMode()` then casts that enum to int and calls `SetRXAMode(m_channelId, val)` (`src/core/RxChannel.cpp:359-370`). The live mode-change handler does the same when an `RxChannel` exists (`src/models/RadioModel.cpp:5202-5206`). If a session restores in `RADE_U`/`RADE_L`, or if we fix RADE RX by keeping the WDSP channel alive, WDSP can see raw mode values 12/13. I think the fix needs a WDSP-facing mapper: keep the slice/UI mode as `RADE_U`/`RADE_L`, but send `USB`/`LSB` (or whatever the intended demod front-end is) to `RxChannel::setMode()`/`onModeChanged()`.

- **P1: `RxDspWorker` can invoke a freed `RadeChannel` across threads.** The worker stores a raw atomic `RadeChannel*` (`src/models/RxDspWorker.cpp:112-114`), loads it during batch processing (`src/models/RxDspWorker.cpp:192-204`), then calls `QMetaObject::invokeMethod(radeCh, "processIq", Qt::QueuedConnection, ...)` (`src/models/RxDspWorker.cpp:247-250`). Teardown only clears the pointer from a `QObject::destroyed` lambda (`src/models/RadioModel.cpp:1974-1981`) while `WdspEngine::destroyRadeChannel()` stops and erases/deletes the object (`src/core/WdspEngine.cpp:519-525`). The atomic protects the pointer value, not QObject lifetime; if teardown lands after the load and before `invokeMethod`, the receiver pointer can already be stale. I would rather see a queued signal owned by the worker and connected to `RadeChannel::processIq`, or teardown serialized with the DSP worker before erasing the channel. `QPointer` may help detection, but the ownership/threading contract still needs to be explicit.

- **P1: the test suite does not compile when tests are actually enabled, and CI currently masks that.** Locally, `cmake -B build-review-tests -G Ninja -DCMAKE_BUILD_TYPE=RelWithDebInfo -DNEREUS_GPU_SPECTRUM=OFF -DNEREUS_BUILD_TESTS=ON` configured, but `cmake --build build-review-tests` failed in `tests/tst_mainwindow_tools_spot_hub.cpp`. The test still constructs `SpotHubDialog` as `(SpotModel*, DxccColorProvider*, parent)` at `tests/tst_mainwindow_tools_spot_hub.cpp:100-110`, while the constructor now requires `SpotTableModel*` before `DxccColorProvider*` (`src/gui/SpotHubDialog.h:241-251`). CI does not catch this because tests are opt-in (`CMakeLists.txt:1140-1150`), the Linux/macOS/Windows configure steps do not pass `-DNEREUS_BUILD_TESTS=ON` (`.github/workflows/ci.yml:289-315`), and the `ctest` step uses `--no-tests=ignore` (`.github/workflows/ci.yml:332-335`). At minimum, one CI job should build tests, and the constructor test should be updated with the new `SpotTableModel` dependency.

- **P1/P2: some RADE tests encode the broken channel-lifetime contract.** `tests/tst_slice_model_rade_swap.cpp:106-139` asserts `setDspMode(RADE_U/RADE_L)` must destroy the WDSP `RxChannel`. That conflicts with the new RX pipeline comments and behavior, where WDSP must stay alive to produce S-meter/spectrum/AGC and to feed RADE post-SSB-demod audio. These tests should be rewritten around the actual intended graph: WDSP `RxChannel` remains alive, `RadeChannel` exists/runs for the slice, and WDSP-facing mode maps to USB/LSB while slice mode remains RADE_U/L.

- **P2: PSK Reporter Start can validate the new identity but still send stale/empty identity.** The PSK tab validates call/grid and writes flat keys `PskReporterCallsign` / `PskReporterGrid` (`src/gui/SpotHubDialog.cpp:2078-2089`), then emits `pskStartRequested`. The connected handler only sets the auto-send interval (`src/gui/MainWindow.cpp:5194-5199`); it does not call `PskReporterClient::setIdentity()`. Meanwhile `RadioModel` initializes/restores from slash keys `PskReporter/Callsign` / `PskReporter/GridSquare` (`src/models/RadioModel.cpp:1000-1004`, `src/models/RadioModel.cpp:1575-1582`), not the flat keys the PSK tab writes. So a user who only fills the PSK tab can click Start after passing validation and still have the live client encode stale or empty receiver fields (`src/core/PskReporterClient.cpp:383-384`). I would unify the key family or read both, and apply `setIdentity(call, gridSquare, version)` before starting.

- **P2: Save & Propagate recomputes FreeDV distances/headings but open reporter rows are not notified.** The identity save path emits `identitySaved` (`src/gui/SpotHubDialog.cpp:523`), and `MainWindow` forwards the new grid into `FreeDVStationModel::setOurGridSquare()` (`src/gui/MainWindow.cpp:5214-5223`). That method re-stamps every existing station (`src/models/FreeDVStationModel.cpp:210-220`) but emits no `stationUpdated`, reset, or dedicated data-changed signal. The open `FreeDVReporterDialog` only repaints from station added/updated/removed/cleared signals (`src/gui/FreeDVReporterDialog.cpp:772-780`). Result: existing rows keep stale Distance/Hdg until each station receives another network update or the dialog is rebuilt.

- **P2/build reliability: the RADE Opus build is network-dependent and large.** `third_party/rade/cmake/BuildOpus.cmake:17` enables `--enable-osce --enable-dred`; in my local test-enabled build this triggered a build-time download from `media.xiph.org` of a ~175 MB Opus model archive. Combined with the earlier `./autogen.sh && ./configure` portability issue, this makes CI slow, brittle, and not hermetic. If those model assets are required, they should be pinned/cached explicitly or the optional model features should be disabled for normal CI builds.

### Areas I would pound on before merge

- RADE restore in `RADE_U`/`RADE_L`: assert WDSP-facing mode is USB/LSB, WDSP `RxChannel` remains alive, `RadeChannel` exists, and RADE receives post-WDSP demod audio.
- Rapid mode toggles: USB -> RADE_U -> RADE_L -> USB while the DSP worker is processing, ideally under ASAN/TSAN or at least with a teardown/recreate stress test.
- PSK tab-only identity flow: enter call/grid on the PSK tab, click Start, assert the live `PskReporterClient` identity and emitted datagram receiver fields use those values; repeat after restart with AutoStart true.
- FreeDV identity save while the Reporter dialog is open: seed a station, change our grid, and assert Distance/Hdg cells refresh immediately.
- CI: add a test-enabled job and remove the possibility that `ctest` silently passes with zero tests.

PR #244, opened 2026-05-14
fix(persistence): atomic settings writes + .bak rotation + corrupt-file preservation (#241)

### LINKED ISSUE #241: Settings file vulnerable to filesystem-level corruption after system events; silent reset to defaults on next launch destroys the only copy
### Bug Description

# Summary

After a hard system event (a hard hang caused by [other open issue]), NereusSDR's `NereusSDR.settings` file was found to have its leading **114,688 bytes (28 × 4 KB sectors) zeroed**, with valid XML content resuming partway through what was originally a tag value. This is consistent with NTFS journal rollback zeroing data sectors during recovery, not just an interrupted file write.

On the next launch, NereusSDR detected the parse failure, logged `WRN: XML parse error in settings: "Start tag expected."`, silently loaded zero settings, and overwrote the corrupted file with factory-default content within minutes. **No backup, no warning, no recovery path. The user's configuration is gone.**

A subsequent investigation confirmed that NereusSDR persists state **only** in `%LOCALAPPDATA%\NereusSDR\NereusSDR.settings` — there is no secondary Windows Registry presence, no `.bak` file, no alternate AppData location. When that one file is corrupted, there is no way for the user (or the application) to recover.

Compounding the issue, factory-default settings include populated TX profile templates (`D-104`, `ESSB`, `PR40`, `PR781`, etc.) and per-band configuration scaffolding. This creates a misleading appearance of "settings are mostly there" when in reality the user's customizations are entirely gone — the populated UI is showing factory presets rather than restored user state. I personally spent significant time attempting partial-recovery surgery on the file before realizing that what I was seeing in the restored UI was templates, not my data.

# Versions / Environment

| Item | Value |
|---|---|
| NereusSDR version | **0.4.0** |
| OS | Windows 11 25H2 |
| Filesystem | NTFS on a WD Blue SA510 SSD |
| Settings file path | `%LOCALAPPDATA%\NereusSDR\NereusSDR.settings` |
| Settings file size when healthy | ~608 KB |
| Settings file format | Custom hybrid: XML element tags with pipe-delimited (`\|`) values inside |

# What I observed

## Settings file state after a hard hang

Raw byte inspection of the post-hang settings file:

```
File size:              608,457 bytes
First 20 bytes (hex):   00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Leading null bytes:     114,688 (exactly 28 × 4 KB sectors)

First non-null content (UTF-8):
|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0|0.0...
```

The exact null-byte count (`114,688 = 28 × 4096`) is the clearest signal: this is sector-aligned data zeroing, almost certainly performed by NTFS during journal recovery after a force-reset, not by NereusSDR itself.

The closing `</NereusSDR>` tag at the file's end was preserved, but everything alphabetically before "hardware" was destroyed (UI layout state, band plan, container definitions, and presumably much more — we can't know exactly what was lost because that data is the part that was zeroed).

## NereusSDR's reaction on next launch

From the next session's log:

```
[22:43:36.796] WRN: XML parse error in settings: "Start tag expected."
[22:43:36.798] DBG: Loaded 0 settings, 0 station settings; 0 saved-radio keys across 0 MAC(s)
[22:43:36.798] DBG: Migrating settings to schema v3 (NereusSDR v0.3.0)
[22:43:36.798] DBG: Settings migration to schema v3 complete
[22:43:36.798] DBG: Migrating settings to schema v4 (averaging math fix)
[22:43:36.798] DBG: Settings migration to schema v4 complete
[22:43:36.798] DBG: Migrating settings to schema v5 (DSP-Options RX/TX split)
[22:43:36.798] DBG: Settings migration to schema v5 complete
```

NereusSDR proceeded with zero loaded settings and started normal operation, including running schema migrations on the empty state. Within minutes it wrote a new factory-defaults file over the corrupt one, destroying the only remaining (partially-recoverable) copy.

**No warning was shown to the user.** The application UI presented as a configured installation, with no indication that settings had been lost or that a corrupt file had just been overwritten. The presence of factory TX profile templates (`D-104`, `ESSB`, etc.) in defaults makes this especially deceptive — the app looks "configured" when in fact it's at zero.

## No recovery options

Confirmed during investigation:

- **No registry presence:** `HKCU\Software` recursive search returned zero NereusSDR-related entries
- **No alternate AppData files:** Recursive search of `%APPDATA%` and `%LOCALAPPDATA%` for `nereus*` returned nothing besides the settings file itself, the log files, and an `impulse_cache.bin` / `wdspWisdom00`
- **No automatic backups:** No `.bak`, `.old`, `.previous`, or rotated versions in the settings directory
- **Volume Shadow Copy contained the same corrupt file:** The single available snapshot (taken earlier on the day of the hang) already had the post-corruption file, suggesting corruption can occur well before any VSS snapshot is taken

There is genuinely no way to recover a user's configuration once `NereusSDR.settings` is corrupted, even with full filesystem and system-restore tooling.

# Why this is severe

1. **Permanent user data loss with no warning.** The user has no way to know their settings were corrupt and are about to be permanently destroyed.
2. **The recovery from a hung session destroys evidence.** Even if the user starts inspecting the file themselves after a crash, NereusSDR will already have overwritten it by the time they look.
3. **Factory templates create a misleading "recovered" appearance.** Users may not realize their customizations are gone until they try to operate the radio and find their TX EQ, band memories, and profile selections don't match what they had configured.
4. **Filesystem-level corruption suggests non-atomic writes.** Sector-aligned zeroing means NereusSDR's settings writes are not crash-safe. Other apps on the same machine (Thetis, browsers, Office, etc.) handle equivalent unclean shutdowns without losing their config, so this is a NereusSDR persistence design issue, not a Windows / NTFS limitation.
5. **Reproducible — any future unclean shutdown will hit this same path.** Power loss, BSOD from any cause, unrelated app hanging the system, even a forced reboot during an unrelated update — all would put NereusSDR's settings file at risk.

# Suggested fixes

In order of increasing implementation effort. Each one alone significantly improves the situation; all of them together would make settings essentially indestructible.

## 1. Preserve the corrupt file before overwriting (smallest change, largest user benefit)

When NereusSDR detects `XML parse error in settings`, do not silently overwrite. Instead:

- Rename the corrupt file to `NereusSDR.settings.corrupt-YYYYMMDD-HHMMSS`
- Log the rename so the user can find it later
- Optionally show a one-time UI notification ("Previous settings could not be loaded. The corrupt file has been preserved as `<path>` in case you want to attempt recovery.")
- THEN start with defaults

This alone would give users a chance to recover via outside tools, professional help, or developer assistance. Cost: roughly ten lines of code.

## 2. Keep a rotating `.bak` of the previous good save

On each successful save:

- Before writing the new file, copy current `NereusSDR.settings` to `NereusSDR.settings.bak`
- Optionally keep `NereusSDR.settings.bak.1` and `.bak.2` for two-deep history

On startup, if `NereusSDR.settings` fails to parse, **try `NereusSDR.settings.bak` automatically before falling back to defaults.**

Cost: ~30 lines of code. Cost to user: zero risk, ~600 KB of disk per backup.

## 3. Atomic settings writes (the real fix)

Settings should be written via the canonical atomic-write pattern:

1. Write new content to `NereusSDR.settings.tmp`
2. `fsync` the temp file
3. Rename `NereusSDR.settings.tmp` over `NereusSDR.settings`

NTFS treats `MoveFileEx(MOVEFILE_REPLACE_EXISTING)` as atomic at the metadata level. The file is either entirely the old version or entirely the new version, never a partial overlay. **No NTFS journal rollback path will zero sectors of a file that wasn't being directly written to**, so the leading-114KB-of-nulls scenario becomes impossible.

Qt provides this functionality directly via `QSaveFile`, which handles temp-write + fsync + atomic-rename in a few lines. If NereusSDR is currently using `QFile` for settings writes, switching to `QSaveFile` is a small, localized change with no API rework.

Cost: small refactor at the save call site(s).

## 4. Validate file structure on read; detect partial corruption

Before parsing the XML, check for:

- File starts with `\x00` bytes (clearly corrupt)
- File is missing closing root tag
- File size suspiciously different from last known good size (could be a soft check)

If any of these fire, treat as corruption and apply (1) and (2) above to preserve the corrupt file and try the backup.

# What I can additionally provide

- The actual corrupted file (608 KB, 114,688 leading nulls) for forensic inspection
- The "defaults file" NereusSDR wrote after detecting the corruption
- The Volume Shadow Copy version of the file (same corruption, different filesystem snapshot)
- Full session logs from the launch where corruption was detected and silently overwritten

# Cross-reference

This issue was discovered in the course of investigating [https://github.com/boydsoftprez/NereusSDR/issues/240]. The hang itself causes the system event that triggers the NTFS-level corruption, but **the silent destroy-on-load behavior is an independent bug that would worsen any other crash scenario** — power loss, BSOD from a different cause, user-forced restart of a hung process, etc. The fixes proposed here would help any user who experiences any kind of unclean shutdown, not just users hitting the underlying hang bug.

Even when the hang is fixed, atomic writes and backups should still be in place. They cost very little and benefit every user on every platform.

### Steps to Reproduce

1. Cause NereusSDR to hard hang (See [https://github.com/https://github.com/boydsoftprez/NereusSDR/issues/240])
2. Reboot 
3. Restart NereusSDR and it will resort to default settings 

### Radio Model & Protocol

HL2

### Radio Firmware Version

74

### NereusSDR Version

4.0

### Operating System

Windows 11

### Relevant Logs

```text

```

boydsoftprez:
## Summary

- Closes #241 — settings file vulnerable to NTFS journal-rollback corruption with no recovery path.
- Switches `AppSettings::save()` from plain `QFile` to **`QSaveFile`** (write-temp + fsync + atomic rename). The leading-NUL-sectors shape reported in #241 becomes structurally impossible.
- Adds **`.bak` rotation** on every save (one-deep history via `<file>.bak.tmp` staging + atomic rename).
- `load()` now detects two corruption shapes (leading-NUL bytes, mid-stream XML parse error) and **preserves the corrupt file as `<file>.corrupt-YYYYMMDD-HHMMSS`** before doing anything else.
- After preservation, `load()` **auto-recovers from `<file>.bak`** when it parses cleanly. The next save rewrites both files.
- Three new query-only accessors (`wasCorruptedOnLoad()` / `preservedCorruptFilePath()` / `recoveredFromBackup()`) feed a future MainWindow UI notification.
- 8 new test cases in `tst_app_settings_corruption.cpp`. Full ctest run: 447/448 green (the only failure, `tst_app_settings_profile`, is pre-existing on main).

This is the structural fix the reporter requested — all four mitigations in their issue write-up are implemented. The only deferred follow-up is a startup UI notification ("Previous settings could not be loaded; preserved as ..."); the accessors are in place, MainWindow can pick them up in a separate small change.

## Test plan

- [x] `ctest -R tst_app_settings_corruption` passes (8/8)
- [x] Full `ctest` run: 447/448 (pre-existing unrelated failure on main)
- [x] Pre-commit verifier chain (Thetis headers, freedv headers, port classifier, inline cites, compliance inventory) all green
- [ ] Manual bench: connect ANAN-G2, configure a non-default value, kill -9 the app mid-save, verify next launch reads back the previous good state from `.bak` and logs the `.corrupt-*` path
- [ ] Windows manual bench: install via NSIS, configure, force a BSOD or hold the power button, verify recovery on relaunch
- [ ] Inspect `~/Library/Preferences/NereusSDR/` (macOS) or `%LOCALAPPDATA%\NereusSDR\` (Windows) after a normal session and confirm the directory contains main + `.bak` (no `.bak.tmp` residue, no QSaveFile temp residue)

### DIFF
--- a/src/core/AppSettings.cpp
+++ b/src/core/AppSettings.cpp
@@ -62,10 +62,12 @@
 
 #include "AppSettings.h"
 
+#include <QDateTime>
 #include <QDir>
 #include <QFile>
 #include <QFileInfo>
 #include <QRegularExpression>
+#include <QSaveFile>
 #include <QSet>
 #include <QStandardPaths>
 #include <QXmlStreamReader>
@@ -336,22 +338,61 @@ static QString sanitizeXmlForLoad(const QString& text)
     return out;
 }
 
-void AppSettings::load()
-{
-    QFile file(m_filePath);
-    if (!file.open(QIODevice::ReadOnly | QIODevice::Text)) {
-        qDebug() << "No settings file found at" << m_filePath << "— using defaults";
-        return;
-    }
+// Corruption-aware load helpers (issue #241).
+//
+// readFileForParse: returns the raw XML text only when the file looks
+//                   intact enough to attempt a parse. Returns nullopt for
+//                   "file is missing", "file is unreadable", "file is
+//                   empty", and "file starts with NUL bytes" (the canonical
+//                   shape produced by an NTFS journal rollback over an
+//                   in-flight write — see issue #241).
+//
+// parseSettingsXml: parses sanitized XML into the supplied QMaps and
+//                   returns true only when the parser reaches the end
+//                   without raising hasError(). On a clean failure both
+//                   maps are left empty so the caller can fall through to
+//                   the .bak path without inheriting half-loaded keys.
+namespace {
 
-    // Read the entire file as UTF-8 text and run it through the recovery
-    // sanitizer before parsing. Pre-2026-04-30 builds wrote element names
-    // with literal '+' (and other non-NameChars from Thetis profile names),
-    // which causes QXmlStreamReader to bail mid-file. The sanitizer escapes
-    // those chars structurally; well-formed files pass through unchanged.
-    const QString rawXml      = QString::fromUtf8(file.readAll());
-    const QString sanitizedXml = sanitizeXmlForLoad(rawXml);
+enum class ReadResult {
+    Ok,             // file was readable + non-empty + did not start with NULs
+    Missing,        // file does not exist (first-run path — no corruption)
+    Unreadable,     // file exists but open() failed (treat as corrupt)
+    EmptyOrZeroed,  // file is zero bytes or starts with NUL (treat as corrupt)
+};
+
+ReadResult readFileForParse(const QString& path, QString& outXml)
+{
+    QFileInfo info(path);
+    if (!info.exists()) {
+        return ReadResult::Missing;
+    }
+    QFile file(path);
+    if (!file.open(QIODevice::ReadOnly)) {
+        return ReadResult::Unreadable;
+    }
+    const QByteArray bytes = file.readAll();
     file.close();
+    if (bytes.isEmpty()) {
+        return ReadResult::EmptyOrZeroed;
+    }
+    // NTFS journal rollback over an in-flight write zeroes whole sectors at
+    // the head of the file (issue #241 — 28 × 4 KB = 114,688 leading NULs
+    // observed). A leading NUL byte is never valid for our XML output.
+    if (bytes.at(0) == '\0') {
+        return ReadResult::EmptyOrZeroed;
+    }
+    outXml = QString::fromUtf8(bytes);
+    return ReadResult::Ok;
+}
+
+bool parseSettingsXml(const QString& sanitizedXml,
+                      QMap<QString, QString>& outSettings,
+                      QMap<QString, QString>& outStationSettings,
+                      QString& outStationName)
+{
+    outSettings.clear();
+    outStationSettings.clear();
 
     QXmlStreamReader xml(sanitizedXml);
     QString currentStation;
@@ -361,24 +402,22 @@ void AppSettings::load()
         xml.readNext();
         if (xml.isStartElement()) {
             const QString tag = xml.name().toString();
-            if (tag == "NereusSDR") {
+            if (tag == QStringLiteral("NereusSDR")) {
                 continue;
             }
-            // Check if this is a station section
             if (!inStation && xml.attributes().hasAttribute(QStringLiteral("type"))
                 && xml.attributes().value(QStringLiteral("type")) == QStringLiteral("station")) {
                 inStation = true;
                 currentStation = tag;
-                m_stationName = tag;
+                outStationName = tag;
                 continue;
             }
-            // Read element text as value; decode key back to internal format
             const QString value = xml.readElementText();
             const QString key   = decodeXmlKey(tag);
             if (inStation) {
-                m_stationSettings.insert(key, value);
+                outStationSettings.insert(key, value);
             } else {
-                m_settings.insert(key, value);
+                outSettings.insert(key, value);
             }
         } else if (xml.isEndElement() && inStation) {
             if (xml.name().toString() == currentStation) {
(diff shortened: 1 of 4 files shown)


boydsoftprez:
Post-merge review finding:

- [P2] `AppSettings::load()` treats a missing main settings file as first-run and returns before checking `.bak`. The new corruption path renames the corrupt main away, then only restores `.bak` into memory. If the app crashes, is killed, or `QSaveFile::commit()` fails before a later successful `save()`, the next launch has no main file plus a valid `.bak`, and silently starts from defaults. That keeps a data-loss hole in the recovery flow.

Suggested fix: either restore `.bak` back to `m_filePath` immediately after a clean backup parse, or make the `ReadResult::Missing` branch try `.bak` when it exists. I would also add a regression test for: recover from `.bak`, create a fresh `AppSettings` without saving, then `load()` still recovers the previous values.

Reference: `src/core/AppSettings.cpp`, around the `ReadResult::Missing` branch in `AppSettings::load()`.

boydsoftprez:
Good catch. The corrupt-preserve rename leaves exactly the on-disk shape you described, so a kill between recovery and the next save() falls right back into the original #241 hole.

Fixed in #249. Went with the second suggestion (Missing branch now tries .bak before declaring first-run) so load() stays read-only. Added both test cases you asked for. Also fixed `tst_app_settings_profile`, which had been failing locally on macOS since #238 unified the path resolver. Full ctest is 469/469 on the #249 branch.

J.J. Boyd ~ KG4VCF

PR #253, opened 2026-05-15
Fix VOX "juggle to prime" bench symptom on radio connect

boydsoftprez:
## Summary

- Adds `MoxController::primeWdspState()` to re-emit `voxThresholdRequested` / `voxHangTimeRequested` / `antiVoxGainRequested` after the late-wired TxChannel is connected, fixing the "VOX needs juggling to prime" bench symptom reported 2026-05-14.
- Calls it from the `pushTxProcessingChain` lambda alongside the existing DEXP initial-sync block, so the load-time VOX values reach WDSP on every fresh connect (and on profile change).
- Adds `tst_mox_controller_prime_wdsp_state` (6 cases) covering the three signals individually, all-three-in-one, default-value-matches-persisted, and the reconnect re-prime case.

## Root cause

`TransmitModel::loadFromSettings` runs at [RadioModel.cpp:2631](src/models/RadioModel.cpp) — early in the connect sequence, before the `MoxController` → `TxChannel` signal connects at [RadioModel.cpp:3604/3612/3620](src/models/RadioModel.cpp). The existing `TransmitModel` → `MoxController` connects (wired in `RadioModel` ctor, lines 770/812) fire `setVoxThreshold` / `setVoxHangTime` / `setAntiVoxGain` during the load. Each setter's NaN-sentinel "first-call emit" fires `voxThresholdRequested` etc. — but those emits land in a void receiver because `TxChannel` does not exist yet. After `TxChannel` is wired, the sentinels are already consumed, so a same-value setter call short-circuits at the `qFuzzyCompare` guard inside `recompute*()`. WDSP retains its construction-time defaults until the user moves a slider, which generates a *different* value that passes the guard.

The asymmetric design for `antiVoxTau` ([RadioModel.cpp:5025](src/models/RadioModel.cpp)) and `antiVoxRun` ([RadioModel.cpp:5051](src/models/RadioModel.cpp)) — defer the TM→Mox connect until *after* `TxWorkerThread` is wired plus an explicit re-push at lines 5043/5070 — already avoids this race for those two paths. `primeWdspState` retrofits the equivalent semantic for the three older paths without restructuring their connect ordering.

## Test plan

- [x] `ctest -R "tst_mox_controller|tst_transmit_model_vox|tst_audio_tx_input_mic_boost"` → 13/13 pass on macOS arm64.
- [x] `ctest -R "tst_radio_model_drive_path|tst_radio_model_dexp_routing|tst_radio_model_eq_lev_alc_wiring|tst_radio_model_3m1b_ownership|tst_radio_model_push_tx_mode_and_bandpass"` → 5/5 pass.
- [x] Main `NereusSDR.app` builds and signs clean.
- [ ] Bench: connect to ANAN-G2 (or HL2) with non-default persisted VOX threshold/hold values, enable VOX, confirm gating trips at the persisted threshold *without* moving the slider first.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/src/models/RadioModel.cpp
+++ b/src/models/RadioModel.cpp
@@ -3804,6 +3804,29 @@ void RadioModel::connectToRadio(const RadioInfo& info)
                 m_txChannel->setDexpLowCut(m_transmitModel.dexpLowCutHz());
                 m_txChannel->setDexpHighCut(m_transmitModel.dexpHighCutHz());
                 m_txChannel->setDexpRunSideChannelFilter(m_transmitModel.dexpSideChannelFilterEnabled());
+
+                // ── Bench fix 2026-05-14: re-prime MoxController WDSP signals ──
+                //
+                // loadFromSettings (line 2631) ran BEFORE the MoxController ->
+                // TxChannel connects above (lines 3604/3612/3620), so the
+                // first-call NaN-sentinel emit of voxThresholdRequested /
+                // voxHangTimeRequested / antiVoxGainRequested landed in a
+                // void receiver and the sentinels are now consumed.  Reset
+                // them and re-run the recompute helpers so the load-time
+                // values reach the freshly-wired TxChannel.
+                //
+                // Reported bench symptom: "VOX needs juggling to prime" --
+                // sliders show correct visual position on launch (model
+                // restored from per-MAC AppSettings) but WDSP retains its
+                // construction-time defaults until the user moves a slider.
+                //
+                // antiVoxTau and antiVoxRun are not covered here -- their
+                // TM -> Mox connects are deferred to wireConnectionSignals
+                // (lines 5025/5051) where an explicit re-push already
+                // happens after TxWorkerThread is wired.
+                if (m_moxController) {
+                    m_moxController->primeWdspState();
+                }
             };
 
             // 1. txEqEnabledChanged → setTxEqRunning.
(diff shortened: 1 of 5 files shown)


src/models/RadioModel.cpp
```diff
@@ -3804,6 +3804,29 @@ void RadioModel::connectToRadio(const RadioInfo& info)
...
+                // sliders show correct visual position on launch (model
+                // restored from per-MAC AppSettings) but WDSP retains its
+                // construction-time defaults until the user moves a slider.
+                //
+                // antiVoxTau and antiVoxRun are not covered here -- their
+                // TM -> Mox connects are deferred to wireConnectionSignals
+                // (lines 5025/5051) where an explicit re-push already
+                // happens after TxWorkerThread is wired.
+                if (m_moxController) {
+                    m_moxController->primeWdspState();
```
boydsoftprez: This call is safe for the initial direct sync, but `pushTxProcessingChain` is also reused from the `activeProfileChanged` connection whose receiver is `m_txChannel`. After `m_txChannel->moveToThread(m_txWorker.get())`, that queued lambda runs on the TX worker thread, while `MoxController` is main-thread-owned (QTimers plus the `TransmitModel -> MoxController` setters). `primeWdspState()` mutates the MoxController sentinels and emits signals, so profile changes while connected can cross-call it off-affinity and race the main-thread setters. Can we split the Mox re-prime out of this worker-bound helper, or dispatch it through `m_moxController` with a queued invoke so it always runs on the MoxController thread?

boydsoftprez (commented):
Left one thread-affinity note on the profile-change path.

src/models/RadioModel.cpp
boydsoftprez: Good catch. Moved the prime call out of the lambda to the initial-sync site (right after the direct `pushTxProcessingChain()` invocation, before `moveToThread`), so it's unambiguously main-thread. Profile changes don't need the re-prime — the TM → Mox → TxChannel chain handles per-property updates through `recompute()`'s computed-value guard once the sentinels are primed.

Pushed as 171273d2.

PR #260, opened 2026-05-16
fix(setup): persist step-attenuator enable/value across restart (#259)

### LINKED ISSUE #259: Nereus can't remember step att setting
### Bug Description

I went into options and set both RX1 Enable and RX2 Enable for 5dB. When I closed and reopened Nereus, they were both back to zero and unchecked.

Image

### Steps to Reproduce

Open Nereus. Connect to radio.
Go to Settings, Options and set both RX1 Enable and RX2 Enable to a value greater than zero. 
Exit Settings. 
Exit Nereus. 
Open Nereus. Connect to radio, and open Settings, Options.

### Radio Model & Protocol

ANAN-10E

### Radio Firmware Version

_No response_

### NereusSDR Version

0.5.1

### Operating System

macOS (Intel)

### Relevant Logs

```text

```

boydsoftprez:
## Summary

Closes #259 — RX1 Enable / RX2 Enable and their dB values in
**Setup → General → Options → Step Attenuator** were reverting to
unchecked / 0 dB after closing and reopening Nereus on ANAN-10E.

Three reinforcing root causes:

1. **`saveSettings()` never ran at app close.**
   `RadioModel::teardownConnection()` nulls `m_connection` via
   `teardownWorkerThreadedConnection()` *before* emitting
   `connectionStateChanged(Disconnected)`. By the time
   `MainWindow::onConnectionStateChanged` runs the disconnect branch,
   `m_radioModel->connection()` is already `nullptr` and the
   `if (connection())` guard around the save was always false. Fix:
   call `m_stepAttController->saveSettings(m_lastRadioInfo.macAddress)`
   inside `teardownConnection` itself, mirroring the `TransmitModel`
   persistence pattern just above it.

2. **`loadSettings()` didn't emit `stepAttEnabledChanged`.** Even after
   fix (1), the controller's restored `m_stepAttEnabled` never reached
   the UI — only `attenuationChanged` and `preampModeChanged` were
   emitted. Added the missing emit at the end of `loadSettings`.

3. **`GeneralOptionsPage::connectController` only wired the RX1 spinbox.**
   Both checkboxes and both spinboxes now follow the controller. The
   RX2 toggle is additionally routed through `setStepAttEnabled` so it
   actually persists, matching Thetis's `chkRX2StepAtt ↔
   chkHermesStepAttenuator` mirror on shared-ADC boards
   (`setup.cs:15719-15721 [v2.10.3.13]`). Full per-RX state remains
   Phase 3F (multi-pan) scope.

## Test plan

- [x] New unit tests cover (a) save → reload preserves `m_stepAttEnabled`,
      (b) `loadSettings` emits `stepAttEnabledChanged` on the restored
      value, (c) `loadSettings` emits the signal even on a fresh MAC
      (so the controller's default-true reaches the UI on first run).
- [x] `ctest -j6` passes 471/471 locally.
- [ ] Bench verification on ANAN-10E by issue reporter: set RX1 Enable +
      RX2 Enable to 5 dB, close Nereus, reopen, confirm both controls
      come back checked at 5 dB.

J.J. Boyd ~ KG4VCF

## Update — pre-load clobber + lazy SetupDialog stale state + RX2 hidden

After bench review the original RX1↔RX2 mirror was the wrong call —
Thetis stores RX1 and RX2 independently and only mirrors on user
click when both RXes share an ADC (`setup.cs:15741-15760
[v2.10.3.13]`). The right NereusSDR fix needs independent
`m_stepAttEnabledRx2` / `m_attDbRx2` + an analogous independent
RX2 auto-att path + a click-time mirror — that is a real refactor
and is queued as a follow-up.

For this PR the RX2 step-att row and the Auto Attenuate RX2
groupbox in **Setup → General → Options** are constructed but
`setVisible(false)`. The existing wiring scaffolding stays in place
so the follow-up touches only the controller side.

A second issue surfaced during HL2 bench testing: persistence of
the RX1 dB value was still failing after the f838a370 commit. The
diagnostic trace caught the clobber path:

```
18:31:11.242 saveSettings m_attDb=0      <-- pre-load teardown
18:31:15.854 loadSettings DONE m_attDb=0  <-- reads back the clobber
```

`RadioModel::connectToRadio` calls `teardownConnection()` whenever
`m_connection` already exists (auto-reconnect retry, panel-driven
reconnect, etc). That teardown's `saveSettings` fired before the
matching `loadSettings`, persisting the controller's constructor
defaults over the user's real saved state. The next `loadSettings`
then read back the freshly-clobbered zero.

Fix: `StepAttenuatorController` gains `m_loadedMac`, set by
`loadSettings` and cleared by `markSettingsUnloaded()`.
`saveSettings` short-circuits unless `m_loadedMac == mac`, so a
pre-load teardown is a no-op. `RadioModel::teardownConnection`
calls `markSettingsUnloaded()` after the save so the next connect
must re-load before its own teardown can save.

Third half (the lazy-SetupDialog miss already mentioned above in
the original PR body): the new `GeneralOptionsPage::initFromController()`
runs at construction with `QSignalBlocker` around each set and pulls
the controller's already-restored state into the widgets, so the
page constructed after `loadSettings` shows the right values.

## Bench verification (HL2)

Connect → Setup → General → Options → RX1 Enable ✓, dB = 6 → close →
⌘Q → relaunch → reconnect → open Setup → Options. RX1 Enable still
checked, spinbox at 6. Persisted file: `rx1Enabled=True`,
`rx1Value=6`, `rx1Band/GEN=6`.

New regression tests:

- `tst_step_attenuator_controller`:
  `persistence_saveBeforeLoad_doesNotClobber`,
  `persistence_markUnloaded_reArmsGate`,
  plus updates to the existing persistence cases for the new gate.
- `tst_general_options_page_step_att_init` (new file):
  `initFromController_restoresEnabledAndValue`,
  `initFromController_disabledState`,
  `rx2Row_isHiddenForNow`,
  `autoAttRx2Group_isHiddenForNow`.

Full `ctest -j6`: 472/472 pass.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/src/core/StepAttenuatorController.cpp
+++ b/src/core/StepAttenuatorController.cpp
@@ -1099,8 +1099,16 @@ void StepAttenuatorController::loadSettings(const QString& mac)
     }
 
     // Notify UI of restored values.
+    //
+    // Issue #259 fix: stepAttEnabledChanged is emitted here so the
+    // "RX1 Enable" / "RX2 Enable" checkboxes on Setup → General → Options
+    // pick up the restored value on connect. Without this emit, the
+    // controller's m_stepAttEnabled is correct but the checkbox stays
+    // at its constructor default (unchecked) until the user clicks it
+    // — so persistence appears broken even when the round-trip works.
     emit attenuationChanged(m_attDb);
     emit preampModeChanged(m_preampMode);
+    emit stepAttEnabledChanged(m_stepAttEnabled);
 }
 
 }  // namespace NereusSDR
--- a/src/core/StepAttenuatorController.cpp
+++ b/src/core/StepAttenuatorController.cpp
@@ -948,6 +948,23 @@ void StepAttenuatorController::onDdcMappingChanged()
 
 void StepAttenuatorController::saveSettings(const QString& mac)
 {
+    // Issue #259 gate: refuse to save until loadSettings has populated this
+    // controller for the same MAC. Without this, the teardownConnection
+    // that runs INSIDE RadioModel::connectToRadio (when a previous m_connection
+    // still exists, e.g. auto-reconnect retry) would persist the constructor
+    // defaults — m_attDb=0, m_stepAttEnabled=true — over the user's real
+    // saved state. The bench trace at /tmp/nereus259/run.log shows the
+    // failure mode: 18:31:11.242 save with m_attDb=0 wipes the prior session's
+    // m_attDb=6, followed at 18:31:15.854 by a loadSettings that reads back
+    // the freshly-clobbered zero.
+    //
+    // Empty mac is also rejected — same guard pattern as the
+    // m_transmitModel.persistToSettings block at the matching site in
+    // RadioModel::teardownConnection().
+    if (mac.isEmpty() || m_loadedMac != mac) {
+        return;
+    }
+
     auto& s = AppSettings::instance();
 
     // Step attenuator global config.
@@ -1023,6 +1040,14 @@ void StepAttenuatorController::saveSettings(const QString& mac)
 
 void StepAttenuatorController::loadSettings(const QString& mac)
 {
+    // Issue #259: record the MAC we're loading for so saveSettings can refuse
+    // pre-load writes. Set BEFORE reading values so a synchronous early-
+    // return path (none today, but a future maintainer's guard could
+    // bail mid-load) still leaves the gate in a consistent state — the
+    // save would re-emit our just-read state which is at worst identical
+    // to disk.
+    m_loadedMac = mac;
+
     auto& s = AppSettings::instance();
 
     // Step attenuator global config.
@@ -1101,11 +1126,11 @@ void StepAttenuatorController::loadSettings(const QString& mac)
     // Notify UI of restored values.
     //
     // Issue #259 fix: stepAttEnabledChanged is emitted here so the
-    // "RX1 Enable" / "RX2 Enable" checkboxes on Setup → General → Options
-    // pick up the restored value on connect. Without this emit, the
-    // controller's m_stepAttEnabled is correct but the checkbox stays
-    // at its constructor default (unchecked) until the user clicks it
-    // — so persistence appears broken even when the round-trip works.
+    // "RX1 Enable" checkbox on Setup → General → Options picks up the
+    // restored value on connect. Without this emit, the controller's
+    // m_stepAttEnabled is correct but the checkbox stays at its
+    // constructor default (unchecked) until the user clicks it — so
+    // persistence appears broken even when the round-trip works.
     emit attenuationChanged(m_attDb);
     emit preampModeChanged(m_preampMode);
     emit stepAttEnabledChanged(m_stepAttEnabled);
--- a/src/gui/setup/GeneralOptionsPage.cpp
+++ b/src/gui/setup/GeneralOptionsPage.cpp
@@ -508,9 +508,15 @@ void GeneralOptionsPage::buildStepAttGroup()
             m_ctrl->setStepAttEnabled(on);
         }
     });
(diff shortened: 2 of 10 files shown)


src/core/StepAttenuatorController.cpp
```diff
@@ -948,6 +948,23 @@ void StepAttenuatorController::onDdcMappingChanged()
...
+    // defaults — m_attDb=0, m_stepAttEnabled=true — over the user's real
+    // saved state. The bench trace at /tmp/nereus259/run.log shows the
+    // failure mode: 18:31:11.242 save with m_attDb=0 wipes the prior session's
+    // m_attDb=6, followed at 18:31:15.854 by a loadSettings that reads back
+    // the freshly-clobbered zero.
+    //
+    // Empty mac is also rejected — same guard pattern as the
+    // m_transmitModel.persistToSettings block at the matching site in
+    // RadioModel::teardownConnection().
+    if (mac.isEmpty() || m_loadedMac != mac) {
```
boydsoftprez, line 964: This gate changes `saveSettings()` from a plain serializer into a load-gated operation. That fixes the runtime pre-load teardown path, but existing callers/tests still save a fresh controller without first calling `loadSettings()` (`tst_step_att_on_tx_value::setAttOnTxValue_persistsPerMac` and `tst_board_capability_flag_wiring::adaptiveSetters_remainCallable_whenFlagFalse`). Those saves now silently no-op, so reload sees defaults. Either update those call sites to load first or split the teardown-safe save path from the general save contract.

src/gui/setup/GeneralOptionsPage.cpp
```diff
@@ -645,12 +672,86 @@ void GeneralOptionsPage::connectController()
...
+        QSignalBlocker blk(m_chkRx1StepAttEnable);
+        m_chkRx1StepAttEnable->setChecked(stepOn);
+    }
+    {
+        QSignalBlocker blk(m_spnRx1StepAttValue);
+        m_spnRx1StepAttValue->setValue(attDb);
+    }
+    m_spnRx1StepAttValue->setEnabled(stepOn);
+
+    // Auto-att group — same lazy-construct problem.
```
boydsoftprez: `initFromController()` now fixes the cold-open step-att state and partially restores Auto Att RX1, but it only applies enable and mode. The restored undo/decay checkbox and hold/delay spinbox remain at constructor defaults, so a page opened after `loadSettings()` can show Adaptive mode with the checkbox still unchecked/labeled `Undo`, and the hold control disabled/default even if the controller restored different values. This should pull the remaining auto-att fields too, or leave auto-att out of this initialization path.

boydsoftprez (commented):
Static review pass found two issues worth fixing before merge.

src/core/StepAttenuatorController.cpp
boydsoftprez: Already addressed in follow-up commit 65473e22 — both tests now call `ctrl.loadSettings(kTestMac)` before mutating + saving. That was exactly the CI failure that commit fixed.

J.J. Boyd ~ KG4VCF

src/gui/setup/GeneralOptionsPage.cpp
boydsoftprez: Good catch. Fixed in 8b3c17c7:

- Added three const-noexcept getters to `StepAttenuatorController` (`autoAttUndo`, `adaptiveHoldSeconds`, `autoUndoDelaySec`) for the read side that was previously only set-only.
- Extended `initFromController` to pull all four auto-att fields with the mode-aware Undo↔Decay label flip and the mode-aware hold-vs-delay spinbox value (mirrors `buildAutoAttGroup`'s `cmbMode` and `spnHold` handlers).
- Two new regression tests pin the Adaptive + Classic init paths end-to-end.

J.J. Boyd ~ KG4VCF

Issue #258, opened 2026-05-15
Audio Backend Switch Causes Stack Corruption and System Instability


Chris-W4ORS:
### Bug Description

# NereusSDR Bug Report — Audio Backend Switch Causes Stack Corruption and System Instability

**Date:** May 15, 2026  
**Reported By:** Chris  
**System:** HP Z800 Workstation — Windows 11 Build 10.0.26200.8457  
**NereusSDR Version:** 0.0.0.0 (timestamp: 0x6a06a38f)  
**Radio Hardware:** Hermes Lite 2  

---

## Summary

Switching the audio output device from **Windows Direct to WASAPI** while NereusSDR is actively running and connected to a Hermes Lite 2 radio causes a Qt thread safety violation, resulting in a fatal stack buffer overflow (exception `0xc0000409` in `ucrtbase.dll`). The crash does not shut down the Windows audio stack cleanly, leaving it in a locked and corrupt state that cascades into system-wide instability requiring a full reboot to recover.

---

## Steps to Reproduce

1. Launch NereusSDR and connect to a Hermes Lite 2 radio
2. Allow the audio engine to initialize and begin streaming (ep6 stream established)
3. While actively connected and streaming, open the NereusSDR **Settings** menu
4. Change the audio output device from **Windows Direct** to **WASAPI**
5. NereusSDR immediately freezes — audio engine stops, threading violation occurs, process crashes with stack corruption

---

## Expected Behavior

NereusSDR should safely stop the audio engine before switching audio backends. If switching backends while live is not supported, the option should be greyed out or the user should be warned that a restart is required.

---

## Actual Behavior

NereusSDR crashes immediately with a fatal stack buffer overflow. The Windows audio stack is left in a corrupt/locked state, causing cascading system instability.

---

### Windows Application Event Log

```
Faulting application name: NereusSDR.exe
Version: 0.0.0.0
Timestamp: 0x6a06a38f
Faulting module: ucrtbase.dll
Version: 10.0.26100.8328
Exception code: 0xc0000409 (Stack Buffer Overflow / Security Check Failure)
Fault offset: 0x00000000000a4aee
Process ID: 0x574C
Application path: C:\Program Files\NereusSDR\NereusSDR.exe
Report ID: 853f92ba-5c12-4cdb-abe4-1f90b1f0bfad
Crash time: 5/15/2026 8:11:59 AM
```

### NereusSDR Log — Final Entries Before Crash

[...]

Issue #239, opened 2026-05-13
Will 'connect' even with radio off


funsutton:
### Bug Description

NereusSDR will connect even if my ANAN-10E is not on. It's not connecting to anything, but shows a 'connected' state regardless. 

### Steps to Reproduce

Keep radio off. Then open Nereus and watch it connect anyway. 

### Radio Model & Protocol

ANAN-10E

### NereusSDR Version

0.4.1 dev

### Operating System

macOS (Intel)

### Relevant Logs

```text

```

boydsoftprez:
Thanks for the report. Pinned it down: the app was flipping to "Connected" the moment it sent the first packet to the radio, without waiting to hear anything back. With the radio off, that first packet still went out, so the indicator went green even though nothing was actually flowing.

#242 changes the state machine to sit in "Connecting" until the radio replies with a real I/Q frame. If nothing comes back within 2 seconds, it tears the connection down and drops to "Disconnected" with a typed timeout error, instead of leaving the green pill stuck on.

Fix lands in #242. Will ship in the next post-0.5.0 build once it's merged.

J.J. Boyd ~ KG4VCF

PR #261, opened 2026-05-16
fix(p1,tx): connect log spam + TX teardown moveToThread (#258)

boydsoftprez:
## Summary

Two surgical fixes for the disconnect-time `0xc0000409` stack-buffer-overflow W4ORS hit on Windows 11 / Hermes Lite 2 in #258.  Investigation of the attached app log + Windows event log fault data turned up two real bugs in the disconnect path; the audio-backend-switch the report blames is a red herring (his audio config stayed unchanged across both connect cycles in the log).

- **P1 connect log spam.** `P1RadioConnection::onReadyRead` captured `cs` as a `const` snapshot outside the per-datagram drain loop.  When the radio is already streaming as metis-start lands, the OS UDP buffer holds N ep6 frames at the moment `readyRead` first fires; every iteration of the loop sees `cs == Connecting`, idempotently calls `setState(Connected)`, and re-emits `"P1: Connected, ep6 stream established"`.  Field repro: 181 spurious lines in <10 ms on first connect, which W4ORS reasonably misread as a "reconnect storm" -- but the state machine itself is correct.  Introduced by [78282d76](../commit/78282d76) / #239.
- **TX teardown `moveToThread`.** `RadioModel::teardownConnection` (and `setActiveRxCountLive`) called `m_txChannel->moveToThread(this->thread())` from the main thread AFTER `m_txWorker->stopPump()` already joined the worker.  Qt6 requires the call from the object's CURRENT thread; with m_txChannel still affined to the (just-finished) worker QThread, Qt emits the exact warning seen in W4ORS's log ("Current thread (...) is not the object's thread (...)") and silently aborts the move.  `m_txWorker.reset()` then destroys the QThread while m_txChannel's thread-data still points at it; the dangling internals get walked during m_txChannel's own destruction and surface on Windows as `0xc0000409` in `ucrtbase.dll`.  Has been firing on every disconnect since [52f27bc4](../commit/52f27bc4) (Phase 3M-1c TX pump v3, 2026-05-02) -- usually didn't crash because the destruction order kept things just-in-bounds, but the reporter's specific timing tipped it over.

## Fix

- **Bug A:** drop `const` on `cs`; sync the local snapshot to `Connected` after `setState()` so subsequent iterations of the drain loop skip the branch.
- **Bug B:** move m_txChannel back to the worker's creator thread INSIDE `TxWorkerThread::run()` just before the loop-exit log, on the worker thread itself (which IS m_txChannel's current affinity), so Qt's precondition is satisfied and the move actually succeeds.  The matching call sites in `RadioModel.cpp` are kept as defensive no-ops (Qt early-returns when target == current) with comments pointing at the new run()-side move -- locks in the invariant at the teardown site against a future refactor that breaks the run()-side move.

## Test plan

- [x] `ctest -R tst_p1_loopback_connection` (6 / 6 passing) -- new `firstConnectLogFiresExactlyOncePerBatch` bursts 100 ep6 frames into a single `readyRead` via a new `P1FakeRadio::setAutoStreamEnabled(false)` knob and asserts via `QtMessageHandler` that the "ep6 stream established" log fires exactly once.  Verified to fail at 100 / 1 with the fix reverted.
- [x] `ctest -R tst_tx_worker_thread` (18 / 18 passing) -- new `stopPump_movesTxChannelToCallerThread_noWarning` mirrors the production handoff and asserts BOTH halves of the invariant: `m_txChannel->thread() == main` after stopPump AND zero `"Current thread is not the object's thread"` warnings during teardown.  Verified to fail (ch.thread() returns `TxWorkerThread` instead of main) with the fix reverted.
- [x] `ctest -R tst_cross_thread_teardown` -- still passing, no regression.
- [x] Main app build (`NereusSDR.app` on macOS) still clean.
- [ ] Manual bench on HL2 (W4ORS) or ANAN-G2: connect, disconnect, repeat ×3.  Confirm log shows exactly one `"ep6 stream established"` line per connect and no `QObject::moveToThread` warnings on disconnect.

## Confidence

Bug B is highly suspect for the actual `0xc0000409` crash but the report's four `.dmp` files are all `PAGEDU64` kernel-mode bugcheck dumps from the audio-stack cascade that followed the NereusSDR crash, not the `NereusSDR.exe` usermode crash dump -- so the causal link is inference from app log + Windows event log fault data, not a confirmed stack walk.  If the next build still crashes on the same flow, the WER usermode dump at `%LOCALAPPDATA%\CrashDumps\NereusSDR.exe.*.dmp` would have the actual stack.

Reported by @Chris-W4ORS in #258.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/src/core/P1RadioConnection.cpp
+++ b/src/core/P1RadioConnection.cpp
@@ -2090,7 +2090,18 @@ void P1RadioConnection::onReadyRead()
 {
     if (!m_socket) { return; }
 
-    const ConnectionState cs = state();
+    // Note: not `const` — we update this local snapshot below when the
+    // Connecting -> Connected promotion fires on the first 1032-byte
+    // datagram, so the "ep6 stream established" log line emits exactly
+    // once per readyRead batch instead of once per datagram in the batch.
+    // Without the update, when UDP delivers a burst of ep6 frames in a
+    // single readyRead (common on first-connect because the radio is
+    // already streaming when metis-start lands), every iteration of the
+    // drain loop below sees `cs == Connecting`, calls setState(Connected)
+    // (idempotent), and re-emits the log line.  Issue #258 — observed
+    // 181 spurious "Connected, ep6 stream established" lines in <10 ms
+    // on first connect, misread as a "reconnect storm" in the bug report.
+    ConnectionState cs = state();
     // Only process ep6 data when Connected or Connecting (reconnect attempt).
     if (cs != ConnectionState::Connected && cs != ConnectionState::Connecting) {
         // Drain the socket anyway to avoid buffering.
@@ -2144,6 +2155,9 @@ void P1RadioConnection::onReadyRead()
                 if (m_reconnectTimer) { m_reconnectTimer->stop(); }
                 if (!m_watchdogTimer->isActive()) { m_watchdogTimer->start(); }
                 setState(ConnectionState::Connected);
+                // Sync the local snapshot so subsequent iterations of the
+                // drain loop don't re-enter this branch.  Issue #258.
+                cs = ConnectionState::Connected;
                 if (wasReconnecting) {
                     if (!m_reconnectedLogged) {
                         qCDebug(lcConnection) << "P1: Reconnected, ep6 stream restored";
--- a/src/core/TxWorkerThread.cpp
+++ b/src/core/TxWorkerThread.cpp
@@ -376,6 +376,34 @@ void TxWorkerThread::run()
         dispatchOneBlock();
     }
 
+    // Issue #258 — move m_txChannel back to this worker's creator thread
+    // (i.e. RadioModel's thread = main) BEFORE run() exits.
+    //
+    // Qt6's QObject::moveToThread requires the call to be made from the
+    // object's CURRENT thread.  If RadioModel::teardownConnection (or
+    // setActiveRxCountLive) tries to do this from the main thread AFTER
+    // stopPump returns, m_txChannel's affinity is still THIS (now-finished)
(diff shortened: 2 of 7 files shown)


boydsoftprez:
Review finding:

[P2] `setActiveRxCountLive()` can leave TX pumping disabled after a live RX-count change.

`RadioModel::setActiveRxCountLive()` calls `m_txWorker->stopPump()`, and `stopPump()` closes `m_txMicSource` via `m_micSource->stop()`. Later, Step 6 moves `m_txChannel` back to the worker and calls `startPump()`, but the mic source is still stopped. `TxWorkerThread::run()` immediately exits because `m_micSource->isRunning()` is false, and this PR's new run-side move returns `m_txChannel` to the main thread. Result: after an active RX-count change, TX can silently stop pumping.

Relevant sites:
- `src/models/RadioModel.cpp:8434` stops the pump / mic source.
- `src/models/RadioModel.cpp:8514` restarts the worker without restarting `m_txMicSource`.
- `src/core/TxWorkerThread.cpp:341` exits when the mic source is not running.

Please either restart `m_txMicSource` before Step 6, or split the worker-thread stop from the mic-source shutdown so live reconfiguration can pause the worker without closing its cadence source.

boydsoftprez:
Deferring to follow-up.  The mic-source-not-restarted path in `setActiveRxCountLive` is a real bug but it's only triggered by a live RX-count change (not the disconnect path this PR fixes), and it's pre-existing — this PR doesn't introduce it, just makes the failure mode visible.  Tracking in #269.

J.J. Boyd ~ KG4VCF

Issue #263, opened 2026-05-16
Anan 10e protocol2 Connection problems


nubbyless:
### Bug Description

Hi my Anan 10e 14 bit is running protocol 2 works ok with Thetis and zeussdr but nereus will detect it connect for a few seconds the disconnect and that's it i have tried disabling firewall completely and tried a reinstall of nereus

### Steps to Reproduce

just connect to radio

### Radio Model & Protocol

Anan 10e protocol 2

### NereusSDR Version

10.4 p2

### Operating System

Windows 11

### Relevant Logs

```text

```

boydsoftprez:
Please send a support log. Tools Support Bundle.

nubbyless:
[support-bundle-20260517-195423.zip](https://github.com/user-attachments/files/27937526/support-bundle-20260517-195423.zip)

nubbyless:
i should add this is a 10e original version not the new 10e revised unit

boydsoftprez:
fixed in #267

J.J. Boyd ~ KG4VCF

PR #264, opened 2026-05-17
fix(spots): dedup by lifetime instead of fixed 60s window (#263)

boydsoftprez:
## Summary
- `SpotModel::dedupIndexFor` now reuses the existing index for as long as the prior spot is still alive (`addedMs + lifetimeSeconds * 1000 > nowMs`), instead of the prior fixed 60 s window
- Added `expireOlderThan()` sweeper driven by a 30 s `QTimer` in the ctor so aged spots actually leave `m_spots` (the lifetime field was previously stored but never enforced)
- `removeSpot()` and `clear()` now also drop the dedup cache slot so the next emit mints a fresh index rather than returning a dangling one
- Review follow-up (2026-05-18): added `SpotData::lastSeenMs`, refreshed on every observation. Expiration compares against `lastSeenMs` instead of `addedMs` so a steadily-active station (RBN re-spotting every minute) does not expire mid-stream and bounce the Spot List row

## Why
The 3J-1 closeout dedup used a 60 s window that expired long before the typical 1800 s spot lifetime, so a re-emit of the same callsign at the same kHz from RBN/cluster/PSK Reporter past +60 s minted a fresh index. Because there was no expiration sweeper either, the prior entry stayed in `m_spots` and the panadapter overlay stacked N copies of the same callsign. Operator observed CT3MD x3, TC19TC x3, 8P6PE x3 on 20 m. Reported as #263.

## Test plan
- [x] `tst_spot_model::dedupReusesIndexWhileSpotIsAlive`: same callsign+freq at +90 s under 1800 s lifetime reuses index, emits `spotUpdated`
- [x] `tst_spot_model::dedupMintsFreshIndexAfterSpotExpires`: same callsign+freq at +5 min under 120 s lifetime evicts prior plus mints fresh
- [x] `tst_spot_model::expirationRemovesAgedSpots`: `expireOlderThan` drops only aged spots
- [x] `tst_spot_model::dedupReuseRefreshesLifetime`: re-emit at +100 s under 120 s lifetime survives a +215 s sweep and only expires at +225 s
- [x] All 11 existing spot tests pass (`tst_spot_model`, `tst_spot_table_model`, `tst_spot_overlay_render`, `tst_spothub_*` x3, `tst_radio_model_spot_wiring`, `tst_mainwindow_tools_spot_hub`, `tst_spot_settings_round_trip`, `tst_spot_auto_start`, `tst_spot_collector_parser`)
- [ ] Bench verify on live RBN/cluster: leave SpotHub open for >2 min on a busy band, confirm panadapter overlay shows one label per (callsign, freqBucket)

J.J. Boyd ~ KG4VCF

### DIFF
--- a/src/models/SpotModel.cpp
+++ b/src/models/SpotModel.cpp
@@ -7,6 +7,21 @@
 // (see https://github.com/ten9876/AetherSDR/blob/main/LICENSE).
 //
 // Modification history (NereusSDR):
+//   2026-05-17  J.J. Boyd / KG4VCF  Issue #263 fix.  dedupIndexFor
+//                                    no longer uses a fixed 60 s window
+//                                    (which expired long before the
+//                                    typical 1800 s lifetime, leaving
+//                                    duplicates on the panadapter
+//                                    overlay).  Instead, reuse the
+//                                    existing index for as long as the
+//                                    prior spot is still alive
+//                                    (addedMs + lifetimeSeconds*1000 >
+//                                    nowMs).  Added expireOlderThan()
+//                                    sweeper + 30 s QTimer in the ctor
+//                                    so aged spots actually leave m_spots
+//                                    (previously addedMs + lifetimeSeconds
+//                                    were stored but never enforced).
+//                                    AI tooling: Anthropic Claude Code.
 //   2026-05-11  J.J. Boyd / KG4VCF  Phase 3J-2 Task D1. Initial port.
 //                                    AetherSDR's "AetherSDR" namespace
 //                                    becomes "NereusSDR". applySpotStatus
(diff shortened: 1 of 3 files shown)


boydsoftprez:
Code review note:

`src/models/SpotModel.cpp` line 179: `dedupIndexForAtTime()` now expires against `addedMs`, but `applySpotStatus()` only sets `addedMs` on the first insert. A station re-emitted regularly inside its lifetime will still be removed once `now - firstSeen >= lifetime`, then re-added with a fresh index. That regresses the old "steadily-active station doesn't expire mid-stream" behavior and can cause label/list flicker for active WSJT-X/RBN spots. Please refresh `addedMs` or track `lastSeenMs` when reusing the existing index.

Verification: I fetched PR 264 at `2ff398ef7cbb7c30c2448c3ba4e9f77503f4ec1d` and ran `ctest --test-dir /tmp/nereus-pr264-review-build -R '^tst_spot_model$' --output-on-failure`; `tst_spot_model` passed. Debug test linking needed a temporary `-DCMAKE_EXE_LINKER_FLAGS=-fsanitize=address` workaround because the repo instruments `NereusSDRObjs` with ASAN but only links the main app with ASAN by default.

PR #278, opened 2026-05-22
fix(tci): wire init burst to live RadioModel state

boydsoftprez:
## Summary
- Wire the TCI init burst to live RadioModel state. Closes the Phase 4 Task 4.2 placeholder gap. RF-Kit RF2K-S and other TCI clients now see actual radio frequency / mode / filter / AGC / etc. in the init burst instead of 14.250 MHz USB hardcoded placeholders.
- Source-first port of Thetis sendInitialRadioState (TCIServer.cs:2486-2660 v2.10.3.15). 19 field families wired to existing Q_INVOKABLE shims; 6 new Q_INVOKABLE shims added on RadioModel for fields that needed them (rx2Enabled, monEnabled, tune, powerOn, diglOffset, diguOffset).
- Two latent bugs caught and fixed during the port: (1) rx_volume:* lines were using the LINEAR linearToDbVolume curve but Thetis uses the LOG audioGainToDb curve at TCIServer.cs:4778. New tciAudioGainToDb helper in TciVolume.h restores parity. (2) NR wiring read only rxNrIndex which can disagree with the gate; new wiring collapses both to Thetis "0 = off" semantics (TCIServer.cs:2519-2526).
- CW macro defaults aligned with Thetis null-controller fallback 30/0/30 at TCIServer.cs:6950-6973 (was 25/0/25). Real CwController lands with Phase 3M-2 CW TX.

## Architectural divergences (all documented inline at call sites + shim declarations)
- powerOn maps to isConnected. NereusSDR has no separate Power button vs connection state.
- DIGL/DIGU offsets come from the active slice. Thetis stores them radio-global; NereusSDR stores per-slice for the multi-slice future.
- Per-RX volumes collapse to afLinear. Thetis has 3 sliders (RX0Gain/RX1Gain/RX2Gain); NereusSDR has one until Phase 3F sub-RX modeling.
- TXFreq emits vfoHz(0,0) for the single-RX !VFOBTX path (the only path NereusSDR reaches today). Full TXFreq logic at console.cs:11345-11369 needs VFOBTX/VFOATX/VFOSplit state that lands with Phase 3F multi-pan.

## Test plan
- [x] ctest full suite: 476/476 pass locally.
- [x] tst_tci_init_burst_live_state (21 slots) covers vfo / mode / filter / agc / mox / rit / xit / lock / mute / sql / nr / anf / apf / bin / nf / ctun / split / balance / audio-stream / iq / volume (both curves) / tx_profile / calibration / tune / mon / powerOn / rx2Enabled / digl / digu / cw-macros + a distinct-state regression slot that catches re-introduced placeholders.
- [x] tst_tci_init_burst_smoke, tst_tci_init_burst_typo_divergence, tst_tci_init_burst_golden, tst_tci_matrix_runner, tst_tci_radio_model_shims all still green.
- [x] All verifier scripts pass (Thetis headers, freedv headers, inline cite preservation, author tags, provenance).
- [ ] Bench (this is the original bug repro): connect RF-Kit RF2K-S at 192.168.109.254 to NereusSDR with the radio tuned to 40m USB at 7.150 MHz. Capture the init burst. Confirm dds / vfo / tx_frequency lines all read 7150000 not 14250000, modulation reads USB, and tx_frequency_thetis includes "40m" not "20m".
- [ ] Bench (curve fix verify): connect wscat. Confirm CW lines read 30/0/30 not 25/0/25, volume reads -30.0 dB at default afLinear=50, rx_volume reads -6.02 dB (log curve, not -30.0 linear).

## Out of scope
- Real ANAN-G2 bench capture for the synthetic golden file. File header still marks it SYNTHETIC; replace with a bench capture in a follow-up PR.
- aamix multi-pan TXFreq paths (Phase 3F).
- CwController-backed CW macros (Phase 3M-2).
- Sub-RX gain modeling and per-RX volume separation (Phase 3F).

J.J. Boyd ~ KG4VCF

### DIFF
--- a/src/core/TciProtocol.cpp
+++ b/src/core/TciProtocol.cpp
@@ -23,6 +23,30 @@
 
 namespace NereusSDR {
 
+// From Thetis TCIServer.cs:2260-2279 [v2.10.3.15] -- agcModeToTciMode.
+// Ported verbatim, including the default-falls-to-"normal" branch.  Input
+// is the upper-case enum-style name produced by RadioModel::agcMode()
+// ("OFF" / "LONG" / "SLOW" / "MED" / "FAST" / "CUSTOM" / "FIXD"); output
+// is the lowercase TCI wire token Thetis sends ("off" / "long" / "slow" /
+// "normal" / "fast" / "custom").  Accepts MED / NORMAL / MEDIUM as
+// synonyms for "normal" so client-set values that round-trip through the
+// shim layer come back unchanged (Thetis tciModeToAgcMode at
+// TCIServer.cs:2280-2303 normalises identically).
+QString TciProtocol::tciAgcModeForWire(const QString& enumName)
+{
+    const QString u = enumName.trimmed().toUpper();
+    if (u == QLatin1String("OFF")
+     || u == QLatin1String("FIXD")
+     || u == QLatin1String("FIXED")) { return QStringLiteral("off"); }
+    if (u == QLatin1String("LONG"))   { return QStringLiteral("long"); }
+    if (u == QLatin1String("SLOW"))   { return QStringLiteral("slow"); }
+    if (u == QLatin1String("FAST"))   { return QStringLiteral("fast"); }
+    if (u == QLatin1String("CUSTOM")) { return QStringLiteral("custom"); }
+    // MED / NORMAL / MEDIUM all map to "normal" per agcModeToTciMode +
+    // tciModeToAgcMode round-trip semantics.
+    return QStringLiteral("normal");
+}
+
 TciProtocol::TciProtocol(QObject* radio, QObject* parent)
     : QObject(parent)
     , m_radio(radio)
@@ -141,13 +165,25 @@ void TciProtocol::enqueueLocalBroadcastVfo(int rxIndex, qint64 hz)
         const QString txKey   = QStringLiteral("tx_frequency");
         const QString txFrame = QStringLiteral("tx_frequency:%1;").arg(hz);
         m_vfoCoalescer.update(txKey, txFrame);
-        // bespoke tx_frequency_thetis (Phase 4 Task 4.2 format).  Band label
-        // and bRX2Enabled / VFOBTX gates match the init burst.
+        // bespoke tx_frequency_thetis -- read the SAME rx2Enabled state used
+        // by buildInitialRadioStateLines so the live broadcast doesn't flip
+        // RX2 from true to false between init and the first VFO move (review
+        // P2 #3, 2026-05-22).  VFOBTX still hardcoded false until Phase 3F
+        // multi-pan lands the full TXFreq logic per console.cs:11345-11369
+        // [v2.10.3.15].  Format from sendTXFrequencyChanged at
+        // TCIServer.cs:2249-2254 [v2.10.3.15]: tx_frequency_thetis:hz,band,
+        // rx2en,txvfob.
+        bool rx2en = false;
+        QMetaObject::invokeMethod(m_radio, "rx2Enabled",
+                                  Qt::DirectConnection,
+                                  Q_RETURN_ARG(bool, rx2en));
         const QString band = bandLabel(bandFromFrequency(static_cast<double>(hz)));
         const QString txThetisKey = QStringLiteral("tx_frequency_thetis");
         const QString txThetisFrame =
-            QStringLiteral("tx_frequency_thetis:%1,%2,false,false;")
-                .arg(hz).arg(band);
+            QStringLiteral("tx_frequency_thetis:%1,%2,%3,false;")
+                .arg(hz)
+                .arg(band)
+                .arg(rx2en ? QStringLiteral("true") : QStringLiteral("false"));
         m_vfoCoalescer.update(txThetisKey, txThetisFrame);
     }
 }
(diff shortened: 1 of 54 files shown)


src/core/TciProtocol.cpp
```diff
@@ -199,90 +259,364 @@ QStringList TciProtocol::buildInitialRadioStateLines() const
...
+    const int cwMacrosSpeed = 30;  // WPM (Thetis null-controller default)
+    const int cwMacrosDelay = 0;   // ms  (Thetis null-controller default)
+    const int cwKeyerSpeed  = 30;  // WPM (Thetis null-controller default)
+
+    // Split per-rx -- From Thetis TCIServer.cs:2615-2616 [v2.10.3.15].
+    const bool split[2] = { readBoolPerRx("split", 0), readBoolPerRx("split", 1) };
+
+    // IQ / audio stream config -- From Thetis TCIServer.cs:2642-2647
+    // [v2.10.3.15].
+    const int     iqSampleRate        = readIntGlobal("iqSampleRate");
```
boydsoftprez, line 651: [P1] The init burst now reads IQ/audio/volume values from production `RadioModel`, but production defaults do not match the mock/golden defaults. On a fresh real client connect, `RadioModel` defaults `afLinear=0`, `monLinear=0`, `iqSampleRate=0`, and `audioStreamSampleType="Float32"`, while tests assume `50`, `50`, `192000`, and `float32`. Real clients can receive `iq_samplerate:0;`, muted volume lines, and a non-canonical sample type even though CI passes. Align production defaults or read actual live slice/connection state, and add coverage that exercises production-shaped defaults.

src/core/TciProtocol.cpp
```diff
@@ -345,43 +710,45 @@ QStringList TciProtocol::buildInitialRadioStateLines() const
...
+    lines << buildRxBalanceLine(1, 1, bal[1][1]);
 
     // From Thetis TCIServer.cs:2427-2433 [v2.10.3.13]
-    lines << buildAgcModeLine(0, agcMode);
-    lines << buildAgcModeLine(1, agcMode);
-    lines << buildAgcGainLine(0, agcGain);
-    lines << buildAgcGainLine(1, agcGain);
-    lines << buildRxCtunExLine(0, ctun);
-    lines << buildRxCtunExLine(1, ctun);
+    lines << buildAgcModeLine(0, agcMode[0]);
```
boydsoftprez: [P2] This emits `agcMode()` directly, but production `RadioModel::agcMode()` returns values such as `MED` and `FAST`. The TCI wire format documented in this file expects lower-case tokens, with `MED` mapped to `normal`; the mock test seeds already-normalized strings, so it misses the real init burst emitting frames such as `agc_mode:0,MED;`. Normalize in `TciProtocol` or `RadioModel` and add a production-shaped test.

src/core/TciProtocol.cpp
```diff
@@ -93,6 +94,64 @@ void TciProtocol::drainCoalescedNotifications()
...
+    if (rxIndex == 0) {
+        const QString txKey   = QStringLiteral("tx_frequency");
+        const QString txFrame = QStringLiteral("tx_frequency:%1;").arg(hz);
+        m_vfoCoalescer.update(txKey, txFrame);
+        // bespoke tx_frequency_thetis (Phase 4 Task 4.2 format).  Band label
+        // and bRX2Enabled / VFOBTX gates match the init burst.
+        const QString band = bandLabel(bandFromFrequency(static_cast<double>(hz)));
+        const QString txThetisKey = QStringLiteral("tx_frequency_thetis");
+        const QString txThetisFrame =
+            QStringLiteral("tx_frequency_thetis:%1,%2,false,false;")
```
boydsoftprez: [P2] `enqueueLocalBroadcastVfo` hard-codes `tx_frequency_thetis` as `...,false,false;` for every live VFO broadcast. The init burst reads the real `rx2Enabled` state, so after the first RX1 tune a client can see RX2 flip from `true` in init state to `false` in the live TX-frequency frame. Read/pass the same `rx2Enabled()` state used by `buildInitialRadioStateLines()`.

src/core/TciServer.cpp
```diff
@@ -540,6 +549,198 @@ void TciServer::hookAudioAndIqTaps()
...
+            [this, rxIndex](NereusSDR::DSPMode mode) {
+                const QString modeStr = SliceModel::modeName(mode);
+                m_protocol->enqueueLocalBroadcast(
+                    QStringLiteral("modulation:%1,%2;").arg(rxIndex).arg(modeStr));
+            });
+
+    // ── Filter (rx_filter_band: line) ───────────────────────────────────────
+    // Source: Thetis console.FilterChangedHandlers at TCIServer.cs:6732
+    // [v2.10.3.15] routed to OnFilterChanged.
+    connect(slice, &SliceModel::filterChanged, this,
```
boydsoftprez, line 637: [P2] Broadcasting every `filterChanged` creates transient stale frames for TCI-originated filter changes. `rx_filter_band` handling calls `RadioModel::setFilterBand`, which updates low and high separately, and each setter emits `filterChanged`; clients can receive an intermediate frame like `rx_filter_band:0,400,<oldHigh>;` before the final value, plus the existing final echo from `TciProtocol`. Use the atomic `SliceModel::setFilter()` path or suppress the signal-side echo for TCI-originated updates.

src/core/TciServer.cpp
```diff
@@ -540,6 +549,198 @@ void TciServer::hookAudioAndIqTaps()
...
+        SliceModel* slice = existingSlices.at(i);
+        if (slice) {
+            wireSliceForBroadcast(slice, i);
+        }
+    }
+
+    // Connect once -- new slices added after TciServer construction get wired
+    // via this lambda.  RadioModel::sliceAdded fires after the slice is
+    // pushed into m_slices, so sliceAt(index) returns the live pointer.
+    connect(m_model, &RadioModel::sliceAdded, this, [this](int index) {
```
boydsoftprez, line 587: [P2] This new `sliceAdded` hook is lost across a server stop/start. `stop()` disconnects all `RadioModel -> TciServer` connections, including this one, but `start()` only re-hooks audio/IQ taps. If the TCI server is stopped/restarted before slices are added, later slices never get broadcast wiring. Re-run `hookSliceBroadcasts()` from `start()` or avoid disconnecting this connection in `stop()`.

boydsoftprez (commented):
Thanks for the follow-up commits. I reviewed the full PR again, including the new live-broadcast work. I found five issues that need fixing before this is safe to merge: one production/default-state mismatch in the init burst, one AGC wire-format mismatch, and three regressions introduced by the new live-state broadcast changes.

PR #280, opened 2026-05-24
fix(g2e/ps): source-first per-packet PureSignal pairing + ANAN-G2E HermesC10 board family

boydsoftprez:
## Summary

Adds first-class support for the ANAN-G2E (HermesC10 effective board) on
both Protocol 1 and Protocol 2, and fixes the PureSignal lock failure
reported on that radio.  The PS lock fix mirrors Thetis ChannelMaster
sync.c InboundBlock(id=1) [v2.10.3.15] per-packet pairing: pscc() now
receives both DDC streams from the same UDP packet, eliminating the
189-sample cross-stream drift that the prior independent-rings
architecture allowed under Qt queued-connection scheduling.

## What changed

* **Source-first PS pairing**: new RadioConnection::psPairedIqDataReceived
  signal carrying both PS DDC buffers from the same packet.  P2 emits from
  the multi-stream deinterleave loop; P1 emits from parseEp6Frame.
  PsccPump consumes via new onPsPairedIqData slot, no host ring buffering
  (commit c73ae3eb).

* **ANAN-G2E SKU support**: HardwareProfile + BoardCapabilities row +
  discovery (P1 wire byte 0x14 -> HermesC10), per-board PA telemetry
  scaling, PS DDC config (Hermes-class branch), PsForm live info[] /
  GetPk readout, AlexController per-MAC TX-bypass persistence,
  P2 dither/random RX unblock for HermesC10, plus the bench-fix tail
  landed alongside the PS work (commit b47a557f and predecessors).

* **TwoTone default 50 -> 10** per Thetis setup.designer.cs:62236
  [v2.10.3.13] udTwoToneLevel.DefaultValue=10.

## Test plan

- [x] All 14 PS-family ctest binaries green (tst_pscc_pump_paired with
      6 new sub-tests pinning the paired-call contract, plus
      tst_ps_ddc_indices_per_board, tst_puresignal_coordinator,
      tst_psform, tst_ampview_window, tst_psa_indicator,
      tst_radio_model_puresignal_run_wiring, tst_codec_ps_ddc_config,
      tst_receiver_manager_ps_ddc, tst_p1_puresignal_*,
      tst_radio_model_hpsdr_model_push, tst_ps_feedback_channel)
- [x] G2E bench: PureSignal engaged, calcc state=6, feedbk=143
      (mid AutoAtt target band), cor.cnt climbing without resets,
      AmpView curve clean and monotonic matching the Thetis reference
- [ ] HL2 bench: re-verify PS still locks (P1 path also rewired;
      regression audit shows logic unchanged for P1 HL2/Hermes/HermesII
      but bench confirmation recommended)
- [ ] G2 (Saturn-class) bench: same audit shows shared codec inheritance
      so logic identical to G2E; bench confirmation if available

## Notes

No remote-state migration required.  AppSettings persistence keys land
under hardware/<mac>/pureSignal/autoCalEnabled for the new PS-A persist
behavior; existing users default to false on first launch.

J.J. Boyd ~ KG4VCF


boydsoftprez (commented):
**Findings**

- [P1] `src/models/RadioModel.cpp:390` omits `ANAN_G2E` from the reflected-power scaling group, and `src/models/RadioModel.cpp:436` / `src/models/RadioModel.cpp:466` omit it from PA volts/amps scaling. `handlePaTelemetry()` consumes these helpers, so G2E forward/exciter telemetry is partly fixed, but reflected power/SWR use the wrong default constants and PA current stays `0.0`. The `userAdc0Changed` gates in the connection classes also still omit G2E, so the title-bar voltage path will not light up.

- [P2] `src/core/RadioDiscovery.cpp:102` does not add `HermesC10` to `maxReceiversForBoard()`. Both P1 and P2 discovery fall back through this helper when the reply receiver count is absent or zero, so the new P1 G2E fixture currently discovers as `maxReceivers = 1` even though the board caps say 4. That misconfigures `ReceiverManager` and blocks additional receiver creation on discovered G2E radios.

- [P2] `src/core/P2RadioConnection.cpp:745` now updates sample rate only for DDC0-3, but `src/core/P2RadioConnection.cpp:705` can enable up to the board cap, and OrionMKII/Saturn caps still expose 7 receivers. If DDC4+ is enabled at 192k/384k/etc., CmdRx will keep advertising 48 kHz for that stream while the app-side receiver is configured for the selected rate.

**Checks**

Reviewed PR #280 at `40d7260e`. Ran `git diff --check origin/main...origin/pr/280` clean; did not run the test suite.

PR #281, opened 2026-05-24
Applet visibility menu (☰ + Containers) with 4O3A capability gating

boydsoftprez:
## Summary

- Restores the per-applet show/hide menu that was disabled in `25597df`, now
  driven by a new `AppletVisibilityController` with both a user-pref axis
  (persisted) and a capability-gate axis (runtime). Surfaces in **Containers
  > Applets** (top menu) and a ☰ button embedded in the S-Meter title bar.
- Registers every currently-wired applet (10 total: RX, TX, Phone / CW, RADE,
  VAX, PureSignal, Power Genius, Tuner Genius, TCI Server, TCI Clients).
- Gates Power Genius + Tuner Genius on the 4O3A master toggle: when 4O3A is
  off, both applets disappear from the right-side panel and grey out in both
  menus, live (no app restart). User's check-state preference is preserved
  across the gate flip.

## Notable changes

- **New class `AppletVisibilityController`** (`src/gui/applets/`): registry +
  persistence + two-axis visibility model. ~140 lines + unit tests.
- **`AppletPanelWidget` gains `setAppletVisible(applet, bool)`** — order-
  preserving show/hide (no widget churn, no reparent), plus a `setBannerMenu`
  hook for the ☰ button embedded in the S-Meter title bar.
- **Double-header bug sweep across 10 applets** — `appletTitleBar()` was
  being called internally on top of the host-side wrapper title bar, causing
  visible double headers on Power Genius, Tuner Genius, TCI Clients, RADE,
  TCI Server. PureSignalApplet had this fix already; the comment there
  flagged the others for follow-up. Fixed across all 10 (4 wired + 6 ghost).
- **`AmpApplet` renamed** from "AMP" to "Power Genius" to match the PGXL brand.
- **`RadioModel::fourO3AEnabledChanged(bool)` signal** added (one-line) so the
  visibility wiring can react live without a restart.

## Architecture notes

- The controller exposes three states per applet: `isVisible` (user wants),
  `isAvailable` (capability gate), `isEffectivelyVisible` (the AND of both).
  Three signals: `visibilityChanged`, `availabilityChanged`,
  `effectiveVisibilityChanged`.
- TCI applets (Phase 23) migrated off their legacy `TciApplet_Visible` /
  `ClientChainApplet_Visible` AppSettings keys onto the controller's standard
  `AppletXxxVisible` keys. Legacy keys become orphans on upgrade; existing
  users get TCI applets back to default-visible.
- RADE is registered with `defaultVisible=false` because its visibility is
  mode-driven (auto-shown only when the active slice is DSPMode::RADE_U/_L).
  User menu toggle is a one-shot override until the next mode change.

## Test plan

- [ ] Build clean: \`cmake --build build -j\`
- [ ] Unit tests green: \`ctest -R "applet_visibility|applet_panel_set_visible|applet_panel_gutter"\` (4 tests)
- [ ] Launch app — all 9 default-visible applets render with single headers
- [ ] Open \`Containers > Applets\` — all 10 entries present, RADE unchecked, others checked
- [ ] Click ☰ button on S-Meter title bar — same 10 entries
- [ ] Toggle TX from top menu — TX hides; banner menu's TX checkmark mirrors
- [ ] Toggle TX from ☰ — TX reappears in its original stack position
- [ ] Setup → CAT & Network → 4O3A → turn off — Power Genius + Tuner Genius
      grey out in both menus AND disappear from the panel (live, no restart)
- [ ] Turn 4O3A back on — both applets reappear, menu entries un-grey
- [ ] Quit, relaunch — hidden states from before the quit are restored

J.J. Boyd ~ KG4VCF

### DIFF
--- a/src/gui/MainWindow.cpp
+++ b/src/gui/MainWindow.cpp
@@ -280,6 +280,7 @@ warren@wpratt.com
 #include "applets/AppletPanelWidget.h"
 #include "SMeterWidget.h"            // Task 41 (Phase 3P-II): SMeterWidget header wiring
 #include "applets/AmpApplet.h"
+#include "applets/AppletVisibilityController.h"
 #include "applets/RxApplet.h"
 #include "core/PgxlConnection.h"
 #include "core/TgxlConnection.h"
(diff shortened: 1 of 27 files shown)


src/gui/MainWindow.cpp
```diff
@@ -2511,6 +2504,164 @@ void MainWindow::populateDefaultMeter()
...
+    const bool fourO3AOn = m_radioModel && m_radioModel->fourO3AEnabled();
+    m_appletVis->setAvailable(QStringLiteral("Amp"),   fourO3AOn);
+    m_appletVis->setAvailable(QStringLiteral("Tuner"), fourO3AOn);
+
+    // Apply initial visibility state from the controller (in case
+    // AppSettings already had values from a prior session).
+    // Uses effective visibility (user pref AND available).
+    for (const QString& id : m_appletVis->registeredIds()) {
+        if (auto* a = m_appletsById.value(id, nullptr)) {
+            panel->setAppletVisible(a, m_appletVis->isEffectivelyVisible(id));
```
boydsoftprez, line 2599: RADE gets registered with `defaultVisible=false` and the initial controller pass hides the panel wrapper. The mode-change path later only calls `m_radeApplet->setVisible(isRade)` on the child widget, so after startup the wrapper stays hidden when the user switches into `RADE_U`/`RADE_L`. Please route the RADE mode gate through `m_appletVis` / `AppletPanelWidget::setAppletVisible`, or model RADE mode as an availability axis, so RADE modes actually reveal the applet and keep menu state consistent.

src/gui/MainWindow.cpp
```diff
@@ -2422,39 +2423,31 @@ void MainWindow::populateDefaultMeter()
         m_radioModel->boardCapabilities().hasPureSignal);
 
     // Phase 23: TCI applets — live in Container #0 below the existing applets.
-    // Visibility default from AppSettings (TciApplet_Visible / ClientChainApplet_Visible).
+    // Visibility is now managed by AppletVisibilityController below
+    // (registered as ids "Tci" + "ClientChain", keys AppletTciVisible +
+    // AppletClientChainVisible). Legacy keys TciApplet_Visible /
+    // ClientChainApplet_Visible from earlier versions become orphans on
```
boydsoftprez, line 2429: This says the legacy TCI keys become orphans, but the existing `View > Network Applets` submenu still reads/writes `TciApplet_Visible` / `ClientChainApplet_Visible` and directly toggles `m_tciApplet` / `m_clientChainApplet`. That gives users two independent controls for the same applets, with different persistence keys, and the old path toggles only the child widget rather than the new wrapper. Please remove that submenu or wire it through `m_appletVis` so both menus stay in sync.

src/gui/MainWindow.cpp
boydsoftprez: Fixed in 7bf3f273. RADE now registers with defaultVisible=true and the
dspModeChanged lambda calls `m_appletVis->setAvailable("Rade", isRade)`
instead of touching the inner widget. The availability axis gates the
wrapper, and the menu entry greys out when not in a RADE mode.

J.J. Boyd ~ KG4VCF

src/gui/MainWindow.cpp
boydsoftprez: Fixed in 7bf3f273. Removed the View > Network Applets submenu entirely.
TCI Server + Clients are reachable via Containers > Applets and the
panel's ☰ menu, both wired through `m_appletVis`. CAT + MIDI greyed
placeholders that lived in the same submenu deferred to phases 3K-1 /
3K-3, to be re-added wired through the controller at that time.

J.J. Boyd ~ KG4VCF

boydsoftprez:
Bench surfaced a related issue: the S-Meter was snapping to the 2 kW
PGXL scale even with 4O3A disabled. Root cause was a pre-existing
auto-connect path that ignored the master toggle. Folded the fix
into this PR per scope discussion (10f11a78):

  1. Gate the PGXL/TGXL connectToHost block on
     \`m_radioModel->fourO3AEnabled()\`.
  2. In setFourO3AEnabled(false), tear down any live PGXL/TGXL TCP
     socket and reset m_hasAmplifier + m_ampOperate so the S-Meter
     and TunerApplet revert their power scales to barefoot.

Verified at bench: launch log no longer shows "amplifier create"
outbound when 4O3A is off, and the S-Meter face displays the 120 W
barefoot scale instead of 0/500/1k/1.5k/2k.

J.J. Boyd ~ KG4VCF

PR #286, opened 2026-05-25
Audio ring overrun protection + latency tightening + spectrum perf wins

boydsoftprez:
## Summary

Three coordinated workstreams: (1) audio ring overrun protection +
end-to-end latency tightening, (2) spectrum + waterfall smoothness, and
(3) CPU paint optimizations.

### Audio path
- Drop-oldest ring overrun protection in PortAudioBus paCallback with
  crossfade smoothing on the resume sample
- Drop-oldest + paOutputUnderflow + paOutputOverflow atomic counters,
  queryable via the PortAudioBus public API
- Move I/Q dispatch off the Qt main thread (Lever 2) so GUI paint
  storms can no longer starve the DSP path
- Ring buffer capacity capped at 100 ms (was 1 second), bufferSamples
  default 256 to 128, persistence-load bug fix so the user's saved
  `BufferSamples=64` is no longer clobbered by device-picker callbacks
- Net result: PortAudio output latency 22 ms to 18 ms, ring worst-case
  ceiling 280 ms to 170 ms, typical end-to-end latency ~85 ms

### Spectrum + waterfall
- Display FPS aligned to FFT output FPS (both at 30 Hz default) with
  `DisplaySpectrumFps` persisted across launches
- `setDisplayFps` now also syncs `m_wfUpdatePeriodMs` so the waterfall
  throttle stays in lock-step with the paint period
- GPU overlay auto-dirty rate-limited from 30 Hz to 10 Hz when
  per-frame-changing features (Active Peak Hold, Peak Blobs, Noise
  Floor) are enabled. The cache was being defeated; rebuild work
  cut by ~67%

### CPU paint reductions
- QStaticText label cache for SpectrumWidget dBm and frequency scale
  labels (ports AetherSDR PR #2556 pattern)
- `WA_OpaquePaintEvent` on SMeterWidget, FilterPassbandWidget,
  MeterSlider
- MeterWidget per-binding fuzzy guard so MeterPoller value pushes
  with unchanged WDSP readings short-circuit before triggering paint
- Spot overlay visual-equality guard (ports AetherSDR PR #2474)
- HGauge value-change guards, S-Meter sub-pixel guard, spectrum FPS
  retuning

### Cleanup
- All session-only diagnostic fprintf chains removed. Atomic counters
  remain queryable via public accessors for future support tooling.
- Stale debug `#include` headers pruned.

## Test plan

- [ ] Bench: connect to ANAN-G2 over Ethernet at 192 kHz, confirm
  audio is clean with no crackle for 10+ minutes
- [ ] Bench: waterfall scroll is smooth, no perceptible step pattern
- [ ] Bench: tune VFO, swap modes, change filter widths, UI stays
  responsive, no audio glitches
- [ ] Bench: idle CPU below the pre-PR baseline (post-rate-limit
  ~100%, was ~120%)
- [ ] Settings: open Setup, Display, Spectrum Defaults; move FPS
  slider; confirm it persists across restart
- [ ] Settings: persisted `BufferSamples` value survives device
  picker changes (the bug this PR also fixes)

## Known limitations / follow-ups

- Noise Blanker (NB / NB2) produces an audible metallic artifact when
  enabled. WDSP source matches Thetis byte-for-byte and the algorithm
  is correct; investigation continues in a separate ticket. Operator
  facing recommendation: leave NB off until that ticket lands.
- Further CPU wins require splitting SpectrumWidget overlay into
  static-chrome vs dynamic layers (1-2 day refactor) or moving more
  chrome to QRhi (multi-day refactor). Both are tracked as separate
  Tier 2 / Tier 3 work in the performance design notes; a handoff
  prompt for that work was drafted at the end of the session.
- This PR does not touch NB UI surfaces. Defaults remain Thetis
  faithful per the source-first protocol.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/src/core/ReceiverManager.cpp
+++ b/src/core/ReceiverManager.cpp
@@ -87,6 +87,7 @@ void ReceiverManager::setMaxReceivers(int max)
 
 int ReceiverManager::createReceiver()
 {
+    QMutexLocker locker(&m_routingMutex);
     if (m_receivers.size() >= m_maxReceivers) {
         qCWarning(lcReceiver) << "Cannot create receiver: at maximum" << m_maxReceivers;
         return -1;
@@ -107,6 +108,7 @@ int ReceiverManager::createReceiver()
 
 void ReceiverManager::destroyReceiver(int receiverIndex)
 {
+    QMutexLocker locker(&m_routingMutex);
     if (!m_receivers.contains(receiverIndex)) {
         return;
     }
@@ -124,6 +126,7 @@ void ReceiverManager::destroyReceiver(int receiverIndex)
 
 void ReceiverManager::reset()
 {
+    QMutexLocker locker(&m_routingMutex);
     const int priorCount = m_receivers.size();
     const QList<int> indices = m_receivers.keys();
 
@@ -286,6 +289,15 @@ void ReceiverManager::setAdcForReceiver(int receiverIndex, int adcIndex)
 
 void ReceiverManager::feedIqData(int hwReceiverIndex, const QVector<float>& samples)
 {
+    // Lever 2 (2026-05-24): this function now runs on the Connection thread
+    // via Qt::DirectConnection from RadioConnection::iqDataReceived (see the
+    // wire-up in RadioModel.cpp).  Main-thread writers of m_hwToLogical /
+    // m_receivers (createReceiver / destroyReceiver / reset / rebuildHardwareMapping)
+    // serialize through m_routingMutex; the read here takes the same lock so
+    // the hash structure can not flip mid-lookup.  Hot-path cost: one
+    // uncontended mutex acquire per packet (~100 ns) plus the existing
+    // hash lookups + emit setup.
+    QMutexLocker locker(&m_routingMutex);
     auto it = m_hwToLogical.constFind(hwReceiverIndex);
     if (it == m_hwToLogical.constEnd()) {
         if (!m_firstDropLogged) {
@@ -321,6 +333,7 @@ void ReceiverManager::feedIqData(int hwReceiverIndex, const QVector<float>& samp
 
 void ReceiverManager::rebuildHardwareMapping()
 {
+    QMutexLocker locker(&m_routingMutex);
     m_hwToLogical.clear();
 
     // Assign hardware DDC indices to active receivers.
--- a/src/core/Rf2ksConnection.cpp
+++ b/src/core/Rf2ksConnection.cpp
@@ -186,23 +186,132 @@ void Rf2ksConnection::parseData(const QByteArray& body)
     emit dataUpdated(bandM, freqKHz, status);
 }
 
-// Stubs for polling / control / IO - filled in Task 3+5.
-void Rf2ksConnection::connectToAmp(const QString& host, quint16 port) {
-    m_host = host; m_port = port;
+void Rf2ksConnection::connectToAmp(const QString& host, quint16 port)
+{
+    m_host = host;
+    m_port = port;
+    m_consecutiveFailures = 0;
+    m_reconnectBackoffMs = 1000;
+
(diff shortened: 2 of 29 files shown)


src/core/ReceiverManager.cpp
```diff
@@ -286,6 +289,15 @@ void ReceiverManager::setAdcForReceiver(int receiverIndex, int adcIndex)
...
 {
+    // Lever 2 (2026-05-24): this function now runs on the Connection thread
+    // via Qt::DirectConnection from RadioConnection::iqDataReceived (see the
+    // wire-up in RadioModel.cpp).  Main-thread writers of m_hwToLogical /
+    // m_receivers (createReceiver / destroyReceiver / reset / rebuildHardwareMapping)
+    // serialize through m_routingMutex; the read here takes the same lock so
+    // the hash structure can not flip mid-lookup.  Hot-path cost: one
+    // uncontended mutex acquire per packet (~100 ns) plus the existing
+    // hash lookups + emit setup.
+    QMutexLocker locker(&m_routingMutex);
```
boydsoftprez, line 300: This only protects the read side and a few call paths, but after this PR `feedIqData()` runs on the connection thread while main-thread code still mutates `m_receivers` without holding `m_routingMutex` before calling `rebuildHardwareMapping()` (for example `activateReceiver()`, `deactivateReceiver()`, `setDdcMapping()`, and several setters/readers). Live RX-count changes can still race an incoming I/Q packet and mutate/read the same `QMap` concurrently. The lock needs to cover every access to `m_receivers` / `m_hwToLogical`, or the stream needs to be fully quiesced before those mutations.

src/core/Rf2ksConnection.cpp
```diff
@@ -0,0 +1,346 @@
...
+}
+
+void Rf2ksConnection::scheduleReconnect()
+{
+    m_reconnectAttempts++;
+    // Double first, then schedule - so the member always holds the delay
+    // that was just used.  testCurrentBackoffMs() reads this value and the
+    // test verifies the 1 s / 2 s / 4 s / 8 s ... 60 s sequence.
+    m_reconnectBackoffMs = qMin(m_reconnectBackoffMs * 2, 60000);
+    m_reconnectTimer.singleShot(m_reconnectBackoffMs, this, [this]{
```
boydsoftprez, line 316: `m_reconnectTimer.singleShot(...)` is the static single-shot API, so the member timer is not active and `disconnect()` cannot cancel the queued reconnect. Also, `markPollFailure()` schedules reconnect without stopping `m_pollTimer`, so regular polling continues during the intended backoff and can reconnect immediately. Use an actual single-shot member timer (`setSingleShot`, `start`, connected timeout) and stop normal polling once the connection is marked down.

src/core/audio/PortAudioBus.cpp
```diff
@@ -372,27 +418,117 @@ qint64 PortAudioBus::pull(char* data, qint64 maxBytes) {
...
+        // silence-to-signal transition would otherwise produce.
+        const int channels = self->m_negFormat.channels;
+        float lastL = self->m_lastOutL;
+        float lastR = self->m_lastOutR;
+        int crossfadeRem = self->m_crossfadeFramesRem;
+        if (startCrossfade && crossfadeRem == 0) {
+            crossfadeRem = kCrossfadeFrames;
+        }
+
+        bool wasUnderrun = (r >= w);
```
boydsoftprez, line 480: `wasUnderrun` is local to this callback, so a continuous empty ring increments `m_underrunEvents` on every callback, and when samples resume on a later callback the state starts false and the intended underrun-to-resume crossfade never begins. Persist the underrun state across callbacks (for example a callback-owned member flag) so distinct underrun events and resume smoothing work as documented.

boydsoftprez (commented):
I found three issues that should be addressed before merge. The RF2K-S tests build and pass locally, but the receiver routing/threading and reconnect/crossfade behavior need fixes.

PR #291, opened 2026-05-27
TX audio quality: NB live-rate fix + TX I/Q telemetry + mic native-rate path

boydsoftprez:
## Summary

Four-commit cleanup from a bench session investigating "TX audio has
digital jitter on the air" reported on macOS with the MacBook built-in
mic.

* `fix(rx-nb)` (baa10cd4) Live-propagate sample rate to NB1/NB2.
  Source-first port of Thetis `cmaster.c:464-470 SetXcmInrate
  [v2.10.3.13]`.  Without this, WDSP NB1/NB2 internal time constants
  stay configured for the original sample rate after
  `RadioModel::setSampleRateLive` and produce audible metallic ringing
  at higher rates.  Bench-verified.

* `feat(perf+tx-iq)` (1c4ff406) Three TX-path improvements landed
  together:
  - PerfMonitor TX I/Q underrun + producer-rate counters, MOX-gated,
    surfaced in the perf overlay + 1 Hz `perf:` log line.
  - 20 ms pre-prime cushion on MOX engage (P1 + P2) defending against
    producer/consumer phase skew at TX start.
  - PreciseTimer on the P2 TX I/Q drain QTimer aligning with the
    existing heartbeat-timer precision (CoarseTimer can drift +/-5%
    on macOS).

* `feat(audio-mic)` (8f5ebf98) Open the PortAudio capture stream at
  the device's native rate when it differs from the requested 48 kHz,
  and resample on our side with the existing r8brain wrapper.  Adapts
  Thetis `ivac.c:311-340 paWinWasapiExclusive [v2.10.3.15]` philosophy
  to the macOS host API (CoreAudio AUHAL SRC bypass).  No-op when
  native already matches requested.  Adds `Resampler::processInto` as
  an allocation-free real-time-safe variant.

* `docs(arch)` (85b393f3) Audio rate matching design doc roadmapping
  the full Thetis `rmatch` port for speakers + VAX rate matching as
  follow-up work.

## What this does NOT fix

The original on-air symptom: TX audio with the MacBook built-in mic
still shows the digital jitter listeners reported.  Bench evidence in
this PR isolates it to the built-in-mic path specifically (an external
USB interface is clean on the same code path).  Native rate negotiated
as 48 kHz on both devices, so the AUHAL SRC bypass is not the cause
either.  Most likely remaining suspect: macOS Voice Isolation injecting
variable-latency processing on internal mics.  Confirmation and
CoreAudio-HAL voice-processing bypass deferred to a follow-up PR.

## Test plan

* [x] Build clean (487/488 link OK; existing deepfilter SDK-version
      link warnings unchanged)
* [x] Targeted unit tests pass (8/8):
      `tst_nb_family`, `tst_rxchannel_nb2_polish`,
      `tst_p1_mox_wire`, `tst_p2_tx_iq_wire`,
      `tst_mox_controller_basic`, `tst_mox_controller_phase_signals`,
      `tst_radio_model_mox_hardware_flip`, `tst_p2_regression_freeze`
* [ ] Bench: verify NB metallic ringing at higher sample rates is gone
* [ ] Bench: perf overlay shows \`tx iq underruns\` counter behaves
      sensibly (MOX-gated; 0 at idle; non-zero during the still-broken
      MacBook-mic TX case)
* [ ] Bench: external USB audio interface TX path remains clean

J.J. Boyd ~ KG4VCF


tests/CMakeLists.txt
```diff
@@ -81,8 +81,25 @@ function(nereus_add_test name)
...
+    #   cmake --build build --target all_tests          # the lot
+    #
+    # CI uses `all_tests` followed by `ctest`.  Pre-commit hooks
+    # currently don't invoke ctest (they're verification scripts,
+    # not test runs), so nothing in the existing workflow breaks.
+    #
+    # The aggregate `all_tests` target is created at the bottom of
+    # this file from a global property each nereus_add_test() call
+    # appends to.
+    add_executable(${name} EXCLUDE_FROM_ALL ${name}.cpp TestSandboxInit.cpp ${ARGN})
```
boydsoftprez: [P1] This makes every test executable `EXCLUDE_FROM_ALL`, but the PR's workflow still only runs the default `cmake --build build` before `ctest`. The current Linux Actions failures match the result exactly: CTest sees registered tests, then reports `Unable to find executable` for every shard. Please add a Linux build of the `all_tests` aggregate before `ctest` or otherwise make CI explicitly build the registered test targets.

src/core/Rf2ksConnection.cpp
```diff
@@ -0,0 +1,450 @@
...
+}
+
+void Rf2ksConnection::scheduleReconnect()
+{
+    m_reconnectAttempts++;
+    // Double first, then schedule - so the member always holds the delay
+    // that was just used.  testCurrentBackoffMs() reads this value and the
+    // test verifies the 1 s / 2 s / 4 s / 8 s ... 60 s sequence.
+    m_reconnectBackoffMs = qMin(m_reconnectBackoffMs * 2, 60000);
+    m_reconnectTimer.singleShot(m_reconnectBackoffMs, this, [this]{
```
boydsoftprez: [P1] This `singleShot` is not owned by `m_reconnectTimer`, so `disconnect()` cannot cancel it. A pending reconnect lambda can fire after the user disables RF-Kit or after `RadioModel::teardownPeripherals()`, issue `/info`, and restart polling. Please make the reconnect timer cancellable/owned, and also have teardown call `disconnect()` even when the connection is only attempting/retrying rather than already `isConnected()`.

boydsoftprez (commented):
Thanks for the big bench-driven cleanup here. I found a few blockers before this can land: CI currently cannot execute the Linux test suite after the `EXCLUDE_FROM_ALL` change, RF-Kit reconnect teardown can restart after explicit disconnect, and a couple of RF-Kit settings/control paths do not match the UI contract. Details inline.

src/core/Rf2ksConnection.cpp
```diff
@@ -0,0 +1,450 @@
...
+    url.setScheme(QStringLiteral("http"));
+    url.setHost(m_host);
+    url.setPort(m_port);
+    url.setPath(path);
+    QNetworkRequest req(url);
+    req.setHeader(QNetworkRequest::ContentTypeHeader,
+                  QStringLiteral("application/json"));
+    // See issueGet() for the rationale on disabling HTTP keep-alive.
+    req.setRawHeader("Connection", "close");
+    auto* reply = m_nam->sendCustomRequest(req, "PUT", body);
```
boydsoftprez, line 510: [P2] Successful control replies are routed back through `onReplyFinished()` with the same state paths used by GET polling. The tests' fake amp returns `Content-Length: 0`, so a successful PUT `/operate-mode` can parse an empty JSON document and emit an empty operate mode; `/antennas/active` can similarly emit antenna 0. Please either skip `handleResponse()` for write acknowledgements or tag write replies separately from state-poll replies.

src/gui/setup/RfKitPage.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+    m_portSpin = new QSpinBox(connBox);
+    m_portSpin->setRange(1, 65535);
+    m_portSpin->setValue(m_model
+        ? m_model->peripheralValue(QStringLiteral("RfKit_ManualPort"),
+                                   QStringLiteral("8080")).toInt()
+        : 8080);
+    connFm->addRow(tr("Port:"), m_portSpin);
+
+    m_autoReconnect = new QCheckBox(tr("Auto-reconnect on disconnect"), connBox);
```
boydsoftprez, line 149: [P2] This setting is saved by the page, but production code never reads it: `scheduleReconnect()` always retries, and the connect paths only pass host/port. The same applies to the saved poll interval below, which is not applied to `Rf2ksConnection` when auto-connecting or reconnecting. Please wire these controls into the actual connection behavior, or remove/disable them until they are live.

src/core/P2RadioConnection.cpp
```diff
@@ -987,6 +1058,14 @@ void P2RadioConnection::sendTxIq(const float* iq, int n)
         // release: publishes the float writes above before the count increment
         // is observed by the connection thread's acquire load.
         m_txIqRingCount.fetch_add(2, std::memory_order_release);
+        ++pushedPairs;
+    }
+    if (pushedPairs > 0 && m_mox) {
```
boydsoftprez, line 1085: [P2] `sendTxIq()` runs from the TX/audio producer path, but `m_mox` is documented as connection-thread-only state. Reading it here introduces a data race and can make the new producer-rate telemetry unreliable. Please gate this with an atomic/latched TX state or otherwise keep the read on the connection thread.

boydsoftprez:
Four review blockers addressed. Each has a test that was red before the fix.

**[P1] Reconnect timer was not cancellable.** `scheduleReconnect()` called `m_reconnectTimer.singleShot(...)`, which is the static `QTimer::singleShot` invoked through an instance. It compiles and reads as if it arms that timer, but it creates a detached one-shot, so `m_reconnectTimer.stop()` in `disconnect()` was a no-op. Now owned: `setSingleShot(true)` plus a connect to a new `onReconnectTimeout()` slot. `teardownPeripherals()` also gated the RF-Kit disconnect on `isConnected()`, which is false precisely when a retry is pending, so that gate is removed. `disconnect()` is idempotent.

**[P2] `m_mox` data race.** `sendTxIq()` runs on the TX/audio producer thread — it is the producer side of the SPSC ring, using explicit acquire/release on `m_txIqRingCount` — while `m_mox` was a plain `bool` written by `setMox()` on the connection thread. Now `std::atomic<bool>`, matching `AudioEngine::m_moxActive`. Also added the missing `#include <atomic>`; the header already declared four atomics while relying on a transitive include. No unit test: a data race has no deterministic failing test at this level, so I verified behaviour is unchanged across all 25 `tst_p2_*`/`tst_mox_*` tests instead of writing something that passes either way.

**[P2] Write acks parsed as state.** Writes came back through `onReplyFinished()` → `handleResponse()` on the same paths as GET polling. The amp answers with `Content-Length: 0`, and `QJsonDocument::fromJson("")` yields an empty object, so `PUT /operate-mode` emitted `operateModeUpdated("")` and `PUT /antennas/active` emitted antenna 0 — the reply to the command that just changed a setting clobbered it. Writes now carry an `rfkitIsWrite` property and skip `handleResponse()`, while still feeding `markPollSuccess()` since an ack is still proof of liveness. `RecordingAmpServer` needed no change; it was already reproducing the real behaviour, and the four existing control tests passed only because none of them watched for the spurious emission.

**[P2] Settings saved but never read.** Confirmed by grep: neither `RfKit_AutoReconnect` nor `RfKit_PollIntervalMs` had any reader outside the page that writes them, and `setPollIntervalMs()` had zero production callers. Both controls were inert. Added `setAutoReconnect()` with an early return in `scheduleReconnect()`, and `RadioModel::applyRfKitOperatorSettings()` called before both `connectToAmp()` sites so it covers the Setup toggle and per-MAC auto-connect alike.

Not addressed here, flagged separately: `PgxlConnection.cpp:421` and `TgxlConnection.cpp:544` have the same uncancellable-reconnect shape, and their teardown calls carry the same `isConnected()` gate. Both are pre-existing on `main` rather than introduced by this PR.

J.J. Boyd ~ KG4VCF

boydsoftprez:
@codex review

boydsoftprez:
**Six review findings addressed, plus the merge with `main`.**

**Codex `[P1]` — Initialize resampling state before starting the stream.** Confirmed. `Pa_StartStream()` hands the stream to the host API's audio thread, which can invoke the callback before `m_inputResampler` and `m_resampleScratch` exist. A first callback could see a half-constructed `unique_ptr`, a zero-length scratch, or take the no-resampler branch and push native-rate samples into the ring. `open()` now publishes every callback-visible member before starting. `Pa_StartStream`'s return value is checked too; it was discarded, so a failed start reported an open bus that never produced audio. No unit test: this is a happens-before relationship with the host API's audio thread, not a testable seam. Fixed on reasoning from the PortAudio contract, not reproduced.

**Codex `[P1]` — Preserve the channel layout after capture resampling.** Confirmed, and this one was audible. The callback downmixes to mono while `negotiatedFormat()` kept reporting the device's channel count; `AudioEngine::pullTxMic` strides by that count, so a stereo mic had it reading two samples per frame out of a mono ring. New `reportedCaptureChannels()` reports 1 whenever we downmix, and the callback deinterleaves against a new `m_inputStreamChannels` carrying the real stream layout.

**Codex `[P1]` — Size the downmix buffer for configured callbacks.** Confirmed. `float monoBuf[1024]` with `monoN` clamped to it; the Audio page offers 2048, so that deterministically discarded over half of every callback. Replaced with `m_monoScratch`, preallocated in `open()` from `m_cfg.bufferSamples`. Shortfalls now increment a counter instead of vanishing.

**Codex `[P2]` — Decrement lock statistics only for successful locks.** Confirmed and reproduced. `lockMemory()` returns false on RLIMIT_MEMLOCK, but `FFTEngine`, `SpectrumWidget` and `PortAudioBus` all discard the return and unlock unconditionally; `munlock()` returns 0 for a mapped-but-unlocked range. Successfully-locked ranges are now tracked in a registry and `unlockMemory` returns early when it holds no matching entry. New `tst_memory_lock_accounting` is red against the old code (4 of 7 fail, with `bytesLocked` at `18446744073709289472` and `regionsLocked` negative) and green after.

**Two more found while reviewing the rest of the diff:**

`PerfMonitor::Ring::push()` held a `QMutex`, and `recordAudioFillMs()` is called from `PortAudioBus::paCallback`. The perf overlay's `stats()` takes the same lock on the GUI thread and holds it while summing 120 doubles, so roughly once a second the real-time audio thread could block behind a lower-priority one. That is the rule CLAUDE.md states outright, and it sat in the path of the jitter this branch exists to chase: the instrumentation could produce the artifact it was added to measure. The ring is now lock-free, with the first tests PerfMonitor has had.

The MOX pre-prime in `sendTxIq()` wrote 7680 floats into a 16384-float ring with no capacity check, while every other write in that function checks first. The 5 ms drain keeps the ring near empty, so this has most likely never fired, but a fast MOX off/on cycle is enough and the failure mode is garbled I/Q on the air. Clamped to available space.

**Merge with `main`.** Four conflicts, all in test/CI build plumbing where both sides independently solved the same `EXCLUDE_FROM_ALL` problem. Took `main`'s `$<TARGET_OBJECTS:nereus_test_sandbox>` form and kept both sides' additive tail.

26 affected tests green locally alongside the 19 new ones. The bench rows in the test plan above remain unrun.

J.J. Boyd ~ KG4VCF

boydsoftprez:
@codex review

boydsoftprez:
**All five findings from the re-review addressed.** Three commits on top of `23a37742`.

**`[P1]` — Clamp the P1 pre-prime to available ring space.** Correct, and my miss: I clamped this exact shape in the P2 path and never checked P1 for it. Same unguarded `fetch_add` of a 960-sample cushion into a 4032-sample ring, with every other write in `sendTxIq()` checking the count first. Clamped to free space. Two tests, verified red-then-green: `preprimeIsClampedToAvailableRingSpace` fails against the pre-fix code, and `preprimeDeliversFullCushionOnEmptyRing` pins the normal 960-sample cushion so a clamp that over-corrects fails loudly rather than passing quietly.

**`[P2]` — Stop RF-Kit polling after a detected disconnect.** Confirmed. `markPollFailure()` declared the link down and scheduled a retry but left `m_pollTimer` running, so the poller kept hammering a dead endpoint every few hundred ms while the backoff stretched to 60 s. The backoff was decorative. Stopped on the down transition; `onReconnectTimeout()` restarts it when a probe succeeds.

**`[P2]` — Reset RF-Kit operate state on disconnect.** Confirmed, and it had two halves. `isAnyExternalAmpInOperate()` read a cache with no connection check, AND nothing ever published the OPERATE to not-OPERATE transition, so `m_lastRfKitInOperate` stayed latched true and no consumer heard otherwise. Fixed both: the predicate now requires `isConnected()`, and `disconnected()` publishes the transition.

**`[P2]` — Ignore RF-Kit replies after explicit disconnect.** Confirmed. Fixed with a session generation bumped by both `connectToAmp()` and `disconnect()`; replies carry the generation they were issued under and stale ones are dropped outright. Dropped rather than merely ignored for state: counting a stale reply as a success is precisely what revived the link, and counting it as a failure would charge the new session for the old one's error.

**`[P2]` — Clear the current peripheral scope while offline.** Confirmed. Worth noting the header already documented the behaviour you asked for ("returns `m_lastRadioInfo.macAddress` when connected, empty otherwise"), so this restores the documented contract the implementation had drifted from. Gated on `m_connectionState` rather than `isConnected()`, since the latter needs a live `RadioConnection` and the Setup-page tests drive state without one. Verified first that `teardownPeripherals()` reads no MAC, so nothing in the disconnect path depended on the old behaviour.

**On the missing coverage.** The reason none of these lifecycle races had tests is that the existing fakes model a server that is either up for the whole test or never started. New `ControllableAmpServer` can delay a response and stop answering mid-session, which is what made three of these reachable. Every new case was verified red first, the lifecycle ones against a build with the generation guard and poll-timer stop deliberately neutered.

Also fixed on a separate branch: `PgxlConnection` and `TgxlConnection` carried the same uncancellable-reconnect defect as `Rf2ksConnection`, pre-existing on `main`.

The three bench rows in the test plan above remain unrun; this PR is being held for hardware verification before merge.

J.J. Boyd ~ KG4VCF

boydsoftprez:
@codex review

boydsoftprez:
Worked the open review threads. Of the 15 unresolved-and-not-outdated findings, 8 were already fixed by later commits on this branch and the threads just never got marked outdated: the two `PortAudioBus` P1s (resampler state published before `Pa_StartStream`, channel layout via `reportedCaptureChannels`), the write-ack-parsed-as-state P2, stale-session generation tagging, `currentRadioMac()` gating on connection state, RF-Kit operate state requiring a live connection, Auto-reconnect / poll-interval actually being read, and `m_mox` being atomic on P2.

The remaining 7 are fixed in `792aea64`, `7c215569`, `9c3f9f72`.

**Reconnect backoff (P2).** Confirmed, and worse than described. `onReconnectTimeout()` restarted the poll timer as soon as it issued the probe, so with `m_connected` still false every later failure fell through `markPollFailure()`'s `>= 3 && m_connected` guard: the poller was never stopped again and no further reconnect was ever armed. The backoff froze where it was while a dead amp got hammered at full cadence. Moving the restart to the connected transition alone would have made it worse (a failed probe would then never retry at all), so `markPollFailure()` now reschedules when a probe fails while already down. New regression test walks connected to down to two failed probes; verified it fails on the old code at the backoff-growth assertion and passes on the new.

**Save not applying operator settings (P2).** Confirmed. `applyRfKitOperatorSettings()` existed but nothing in `RfKitPage` called it.

**External antenna lighting internal buttons (P2).** Confirmed. Matching on `number` alone with no `type` check.

**RF-Kit telemetry on a cross-vendor gate (P2).** Confirmed. Added `RadioModel::isRfKitInOperate()` and gated the feed on it; `isAnyExternalAmpInOperate()` now delegates rather than duplicating the liveness check.

**Uninitialized dynamic overlay (P1).** Confirmed at both the init and resize sites; both now fill transparent before the lock.

**Static overlay not unlocked on release (P2).** Confirmed and fixed. One thing I did not do: `m_waterfall` has what looks like the identical defect in the same function, but it is locked where it is reallocated on a size change rather than in `initOverlayPipeline()`, so it survives `releaseResources()` still locked and still in use. Unlocking it there would drop a live pin nothing re-establishes. Left alone with a comment saying why.

**P1 producer telemetry (P2).** Confirmed. Gating on `m_mox` to match P2 makes it a cross-thread read, so `P1RadioConnection::m_mox` becomes atomic for the same reason P2's did.

Evidence: full build clean, 18/18 tests green across the RF-Kit, P1 wire and spectrum sets. None of this is bench-verified against a live RF2K-S or an HL2; the reconnect path is covered by the new test, the rest is desk-verified.

Note on the merge state: this branch is behind main and I have deliberately not merged it here, since the TNF integration branch already owns stitching this together.

J.J. Boyd ~ KG4VCF

boydsoftprez:
@codex review

PR #293, opened 2026-05-28
Phase 3F multi-pan multi-slice foundation (sub-epics A-J)

boydsoftprez:
## Summary

Phase 3F multi-pan + multi-slice foundation. 346 commits across sub-epics A through J plus a stabilization pass. Single-PR per the explicit stack-commits directive.

### Headline operator-visible deliverables

- **Multi-pan layout**, 5 templates (Single, Stacked, Side-by-Side, Wide+2, Grid 2x2), float-to-second-monitor.
- **Multi-slice up to maxSlices per SKU** with TxSliceArbiter (single-TX invariant, RF-safe MOX-drop handoff). VFO flag auto-creates per slice.
- **Per-slice control plane (Sub-Epic J).** Every DSP control addresses the slice the operator is working. One rule, two clauses: a control drawn on a flag targets that flag's slice; a control attached to no slice targets the active slice. Enforced by `scripts/verify-no-gui-dsp-access.py`.
- **Per-pan badges**, right-click flag context menu, Alex per-ADC BPF state machine with FilterPolicyDialog, bottom-bar CH 0 / CH 1 indicators.
- **Wideband data path end-to-end**, zoom-past-DDC auto-engages extended mode.
- **Diversity** via Tools > Diversity Dialog, radar plot, 8 per-band memories.
- **Sub-Epic I**: stream/DDC allocation, `SliceStreamAllocator`, per-slice sample rate, PureSignal and diversity stream announcements.

### Landed after the original description was written

- **Sub-Epic I and J** in full, including the per-slice DSP routing this PR originally deferred to 3F-1. Audio, NR, AGC, CTUN and stereo pan are now genuinely per slice.
- **Bench fixes from live G2 and G2E sessions**: mixer readiness barrier, PureSignal MOX crash, the MOX RX-audio gate (implemented, tested, documented as wired, and never actually connected), 10 ms raised-cosine up-slew, per-flag frequency and passband, click-to-tune following the selected flag.
- **Three ADC and filter defects** found by auditing against CT1IQI's review: `P2CodecSaturn` never inherited antenna-driven ADC routing and every test built the wrong codec, so the feature was absent on the radio it was written for; `CodecContext::adcCtrl` was never seeded on Protocol 2 so the diversity pair sat on the same physical input twice; the Alex filter analysis was pinned to ADC0 so chain 1 never got a decision.
- **PR 293 stabilization pass**: TX filter source, TX antenna authority transfer, mixer withdrawal race, extended-view policy versus actual state, RF2KS live-retarget.
- **Chip closeout**: G2E disconnect lockup (merged separately as #306), and the per-slice gaps (NB/SNB per-slice tuning, TCI `rx_volume` and `rx_anf_enable` live broadcast, APF/BIN reaching the DSP instead of a private array).

### Corrections to the previous description

- It said ~110 commits. It is 346.
- It said "Plus PR #291 merge". **#291 is still open and currently conflicting.** Nothing from it is in this branch, including the overlay-texture fix `9c3f9f72`, whose two defects are live here.
- It said per-slice DSP routing was deferred and "the DSP audio you HEAR is still Slice A's". Sub-Epic J shipped it.

### Known limitations

- **Per-slice antenna does not compose with ADC routing on the same band.** Two slices on one band, second set to EXT1: the selection applies and a sibling slice reverts it about 2.6 s later, so neither pan reaches chain 1. Two root causes, both diagnosed and neither fixed. The antenna is stored per band rather than per slice, so the slice on ANT1 clears the other's EXT1 for the whole band. And `SliceStreamAllocator` places slices by frequency alone, so it co-hosts a pair whose antennas have just made co-hosting impossible: one stream is one DDC, and one DDC has one ADC. Full writeup in `docs/architecture/2026-07-30-per-slice-antenna-adc-routing.md`. Row 17 of the G2 matrix covers the same-band case that rows 15 and 16 structurally cannot catch. The fix runs as its own branch.
- `9c3f9f72` (overlay images allocated uninitialised, and a MemoryLock registration leaked on every device or surface recreate) exists only on #291. Both faults scale with pan count here.
- Sub-Epic G landed 8 of its 25 tasks. Diversity beyond the basic engage is deferred.
- Diversity to CT1IQI's specification (DDC n and n+1 with n even, more than one diversity receiver) is not implemented and needs a DivId allocator that does not exist. WDSP `pdiv[]` is a 2-slot array keyed by diversity id and dereferences unallocated state for id >= 2.

## Test plan

- [x] Full suite green: **569/569**, built through the `all_tests` target and then `ctest -j8`. Test binaries are `EXCLUDE_FROM_ALL`, so a bare `ctest` runs stale binaries and reports false green.
- [x] CI green: all 9 checks.
- [x] App launches and connects to ANAN-G2 (Saturn, fw 27) and ANAN-G2E (fw 110).
- [ ] **G2 bench matrix, not run.** `docs/architecture/2026-05-26-phase3f-verification/g2-results.md`, including rows 15 through 17.
- [ ] **Sub-Epic J 8-row bench matrix, not run.**
- [ ] **Two-pan simultaneous audio listen, not run.** This is the test that motivated the epic.
- [ ] Chip `task_16e1289c` (RX audio dies mid-session after TX or TUNE): code landed, bench unconfirmed.

## Risk surface

This reorganises the RX data plane. Single-slice operation has regression coverage, but multi-slice, wideband and diversity are new paths.

**No review of any kind has ever run on this PR.** A Codex review before merge is recommended, and the merge is better gated on the bench matrix than on the suite alone.

J.J. Boyd ~ KG4VCF


CT1IQI:
Attention: the Alex control system as now coded appears to be not aware that systems with dual ADCs like the Anan G2 also have dual input band pass filters. When up to the max. number of DDCs get activated for receive, one must set which ADC serves which DDC. Then the frequencies need to be reviewed that are serviced by each ADC. When for one ADC all are within the reach of one band filter, that band filter needs to become active in the filter chain before that ADC. When frequencies need to be served from two or more band filter ranges, the band filtering needs to be bypassed. This has to be reviewed per ADC, its assigned DDCs, and filter chain. Have not spotted this kind of analysis before settings are made, under the assumption of an Alex setting per ADC.

CT1IQI:
Having continued preparation for a G2 xdma 'protocol' interface, bypassing Protocol2 and UDP packets, have gotten stuck on what to use as the authoritative source for DDC settings and also the introduction of the term 'slice' as some sort of receiver. In ReceiverManager one finds a ReceiverConfig that does not seem to be set completely (did not find any place in Nereus where 'isPureSignalFeedback' or 'isPuresignal' get set. So one wonders about the validity of the DDC information. This 3f pull request adds in P2CodecSaturn some routine 'applyDdcAssignment' which seems not to set DDCs properly.
A 'receiver', with its frequency/mode/bandwidth/used_for_tx settings can require one or two hardware receive channels.
One in case of no diversity, and two in case of diversity. Not just any G2 DDC can be used for diversity, but there can be more than one receivers with diversity set! The G2 FPGA programming is such that in case of diversity reception this requires DDCn and DDC n+1, n being an even number, per diversity receiver. n needs to be set to synchronous and enable, and n+1 to enabled. n to using ADC0 and n+1 to using ADC1, Alex0 and Alex 1 set identical in terms of filtering. When a receiver is not using diversity any DDC can be used that is free. Hence, when assigning DDCs to what I call receivers, first the receivers with diversity need to be accommodated, followed by the rest. It is a choice whether to reserve DDCs for when PureSignal is used. The bandpass filter setting is a routine in itself in that it needs to analyse the ADC0&1 use and the settings of DDC frequency that go with the defined receivers. When one or more receivers are set to diversity, this requires filters to be set equal, but this can be also equally bypass filtering. This depends on whether the other frequencies used via an ADC fit the same filter. In trying to pick up the settings from Nereus to set the FPGA, it is not yet clear to me where this is being analyzed and set properly. Another thing to consider is how to limit the number of receivers. Yes, there is the limit of available DDCs, but how many of those can be used depends on the sample rate set, the available transport bandwidth (per ethernet or xdma), and wdsp's need for processing power. Just as an idea: would it be helpful to introduce a sample rate budget and, when one wants to add receivers to limit them to the available sample rate budget with a warning that to add another receiver one must reduce sample rates of one or more others?

boydsoftprez:
## Sub-Epic I: data-plane completion

35 commits, 86 files, +11890/-490, 5 new test suites. Full suite 545/545 green. **No bench run yet**, so everything below is desk work plus unit coverage, not hardware-verified.

### Why this exists

The G2 bench session on 2026-07-24 found three symptoms: the second pan never animated, Slice C sat on 20 m with no audio, and the radio was never told to enable a second DDC. Triage showed one root cause. Phase 3F landed the multi-pan/multi-slice UI and every supporting component, but the data plane underneath was never connected. Concretely:

* One `FFTEngine` existed, bound at construction to pan 0's `SpectrumWidget`. That was the only call site of `updateSpectrumLinear` in the tree, so a secondary pan had a fully built widget that was never handed a frame.
* `addSlice()` created a `SliceModel` and wired UI signals. No receiver, no DDC, no WDSP channel. `RxDspWorker` hardcoded `rxChannel(0)`.
* `invokeCodecDdcAssignment()` had zero callers, so the entire Sub-Epic B per-board codec layer was unreachable.
* `FFTRouter` was dead: its topology map was written and never read.

### Model

Two concepts that Phase 3F had conflated:

* A **DDC stream** is one hardware DDC plus its `ReceiverManager` receiver, its `FFTEngine`, its panadapter window, and its noise blanker.
* A **slice** is one WDSP RX channel plus a shift offset.

Slices bind to streams **many-to-one**. A slice whose frequency falls inside an active stream's window joins it and is tuned by shift offset. One that fits no active window claims a free DDC. When none is free the add is rejected with a message naming the limit. So Slices A through D on one band cost one DDC, and spreading them across bands costs one each, up to what the SKU has.

Both pools open at connect and are never resized: `userDdcCount` streams and `maxSlices` WDSP channels.

### Source basis

This is not invention. The sharing contract is `ChannelMaster/cmaster.h:75-82 [v2.10.3.15]`, where `struct _rcvr` holds one noise blanker (`ANB panb` / `NOB pnob`) and one panadapter (`run_pan`) per receiver, but `double* audio[cmMAXSubRcvr]` per sub-receiver. Shared I/Q in, independent audio out.

The window rule ports the Multi-RX guard at `console.cs:31913-31925 [v2.10.3.15]`, strict inequalities both sides. The shift offset is `RxChannel::setShiftFrequency`, already in the tree as a port of `RXOsc` at `radio.cs:1409-1420 [v2.10.3.15]`.

Pre-allocation follows Thetis `CreateRadio()` (`cmaster.cs:502-516`, `cmRCVR=5`, `cmSubRCVR=2`) and deskhpsdr's create-all loop (`radio.c:1256-1259 [@f3d857c]`). Neither upstream opens a WDSP channel at runtime.

Three divergences from Thetis are deliberate and recorded in design doc §3: up to `maxSlices` slices per stream rather than Thetis's `cmSubRCVR=2`; every user DDC the SKU exposes rather than the two Thetis uses in practice; and promoting a slice to a free DDC when it leaves its window rather than disabling Multi-RX as Thetis does.

### Notable defects found while implementing

Several of these were found by reading upstream rather than by testing, and would not have been visible until a bench session:

* **`fexchange2` can return without writing its output buffers.** Its body sits inside an exchange-bit check while `*error` is pre-set to 0, so on flush and restart transitions the output legs are untouched. The old per-drain `QVector` allocation was masking this by zero-filling. Hoisting those buffers for the per-slice loop needed an explicit zero-fill to stay correct.
* **The noise blanker aliased co-hosted slices.** `xanbEXTF` / `xnobEXTF` mutate the caller's input legs in place, so once slices share a chunk one slice's blanker corrupts what every later slice sees. Now applied once per stream, matching `_rcvr`.
* **Hermes-class boards would have lost RX on the first VFO turn.** `P2CodecOrionMkII` was selected by `default:` for Hermes / HermesII / HermesC10, and its table puts stream 0 on DDC2, while `primaryRxDdcForBoard` returns 0 for that family. Inert while the codec had no caller; a regression the moment it got one. Fixed by porting the Hermes-class branch as `P2CodecHermes` plus a `P2CodecHermesII` for the nddc=2 case.
* **Slice ids collided with list positions.** `addSlice` stamped the id from the list length and never renumbered, while `AudioEngine` resolved that same integer positionally. After removing a middle slice, one slice inherited another's mute state and VAX channel, and a subsequent add could collide onto a live WDSP channel. Ids are now stable and resolved by id; `sliceAt` was renamed `sliceById` so the ambiguity that produced the bug cannot be re-read.
* **Secondary slice audio was reaching a mixer that dropped it.** `MasterMixer::accumulate` early-returns for unregistered ids and only id 0 was ever registered. All ids are now pre-registered at connect rather than enrolled lazily, because lazy enrollment would rehash the map under the DSP thread.
* **Every VFO tick was rebuilding the hardware map.** `setDdcMapping` had no change gate, and the rebuild re-emits stored frequencies with no CTUN guard, so a CTUN pan drag would snap back on the next VFO click. Gated.

### Findings from an internal review pass, since fixed

A full-branch review after the tasks landed found three more that would each have shown up at the bench as something else:

* **Secondary slice audio was reaching a mixer that dropped it.** `MasterMixer::accumulate` early-returns for unregistered ids, and only id 0 was ever registered. Slices B+ were demodulated correctly and discarded one call later, which is the same symptom that opened this whole thread. All ids are now pre-registered at connect rather than enrolled lazily, because lazy enrollment would rehash the map under the DSP thread.
* **Every VFO tick was rebuilding the hardware map.** `setDdcMapping` had no change gate, and the rebuild re-emits stored frequencies with no CTUN guard, so a CTUN pan drag would snap back on the next VFO click. This one reached single-slice operators. Gated, and `forceHardwareFrequency` now stores the frequency so a rebuild is idempotent.
* **Slice ids collided with list positions.** `addSlice` stamped the id from the list length and never renumbered, while `AudioEngine` resolved that same integer positionally. After a middle removal one slice inherited another's mute state and VAX channel, and a later add could collide onto a live WDSP channel. Ids are stable now and resolved by id; `sliceAt` was renamed `sliceById` so the ambiguity that produced the bug cannot be re-read.

Plus four that would have wasted bench time:

* Stream bindings never reached a freshly constructed `RxDspWorker`, so on reconnect a slice on a non-zero stream demodulated nothing until retuned.
* `kHermesC10` advertised 5 user DDCs against a codec that assigns 4, so a claimable stream could never get an enable bit. Its comment also still described the pre-`P2CodecHermes` DDC2-6 layout.
* PureSignal and diversity collapse the Hermes-class DDC set, and nothing recomputed the assignment on MOX, so slices kept reporting pre-key DDCs. Thetis does drop them too (`console.cs:8448-8456 [v2.10.3.15]` enables DDC0 alone with no `rx2_enabled` addendum on that branch, unlike the ORION class at `:8299-8303`), so the collapse stays and what changed is that it is now announced instead of silent.
* A rejected retune left the VFO reading one frequency while the DSP stayed on another. The frequency now rolls back, and the rejection has its own signal rather than borrowing the add-slice one.

And two more found on a second pass:

* **The anti-VOX cancellation feed fired once per DDC stream rather than once per radio.** With two streams that is double the block rate WDSP DEXP was configured for through `SetAntiVOXSize` / `SetAntiVOXRate`, and half the blocks were a stale repeat, because only slice 0's stream writes the buffer they read. `dexp.c:288-297 [v2.10.3.15]` integrates one block per flag through an IIR calibrated in sample steps, so double delivery halves the effective time constant against wall clock, and `dexp.c:708-715` makes delivery destructive, so a second block arriving before the TXA pump consumed the first silently replaced it. Anti-VOX is what stops speaker audio tripping VOX, so the failure modes were a false trigger or a failure to cancel. Now raised once per radio, gated on whichever stream currently hosts slice 0.
* **The per-slice sample rate control was wired to nothing.** `setStreamSampleRate` existed with tests but had no production caller, while the VFO flag still wrote a `SliceModel` property that `buildStreamConfigsForCodec` no longer reads. The control is now routed to the slice's stream, the rate list is filtered per board, and on P1 the menu says "whole radio" because P1 carries one rate in C&C bank 0.

### What is deferred, deliberately

* All streams share one DSP thread and one FFT thread. Thetis runs a thread per receiver. Splitting is a thread-architecture change and wants maintainer sign-off; a 5-slice 1536 kHz bench is the test that would justify it.
* Co-hosted slices share one noise blanker, which is hardware topology rather than a bug. The UI does not yet say so, so a co-hosted slice's NB toggle looks active while being inert. That is a UX call.
* Multi-slice RADE stays gated to slice 0, matching the v0.5.2 limitation.
* Per-slice sample rate has a working `setStreamSampleRate` with tests, but no production call site yet; the VFO flag still writes the old property.
* `userDdcCount` cannot express the truth as one number per board: `HermesII` is 4 on P1 and 2 on P2, and `kOrionMKII` varies by model as well as protocol. Only the unambiguous `kHermesC10` case was corrected. `kAngelia` and `kOrion` are right for the hardware (Thetis drives both with `nddc = 5`); the gap is that `P1CodecStandard` only implements the HERMES-class branch, so the ORION-class P1 branch is an unported scope gap rather than a wrong capability. Marked `QEXPECT_FAIL` so it XPASSes the day it lands.
* `setActiveSlice` / `TxSliceArbiter::requestHandoff` / `activeSliceChanged` are still positional while MainWindow feeds them slice ids. Pre-existing, and not made worse here, but it is the next thing in that family that will bite. The arbiter's index is persisted, so converting it needs a migration decision.

### Verification status

Full suite 545/545. The bench matrix gains rows 48 through 60 covering DDC sharing, eviction, rate changes, exhaustion messaging, the second pan animating, and a Hermes-class RX regression guard. Rows 1 through 15 became testable for the first time, since before this sub-epic Slices B+ were UI objects with no receiver.

The bench walk to run first is in `docs/architecture/2026-05-26-phase3f-verification/g2-results.md`: four slices sharing one DDC on one band, then one retuned across bands to claim its own DDC and its own pan, then exhaustion, then a single-slice regression check.

J.J. Boyd ~ KG4VCF

boydsoftprez:
Thank you for both of these. They found four defects, three of which are now fixed, and one of them meant a feature did nothing at all on the ANAN-G2. Taking them point by point.

**On the per-ADC filter review (your 2026-05-31 comment).** The analysis you describe does exist, and it was built in response to this comment. You will find it in `AlexController::recomputeBpf` in `src/core/accessories/AlexController.{h,cpp}`, fed by `RadioModel::republishAlexAdcSlices`, with the two independent filter words composed in `src/core/codec/AlexFilterMap.cpp`. It groups slices by the ADC serving them, spans the frequencies on each chain, selects a single band filter if they all fit, and bypasses that chain if they span two or more ranges. It runs before the settings reach the wire, and it produces a per-chain decision with a human-readable reason, which the bottom bar surfaces as text like `BYPASS (multi-band: 40m + 20m)`.

You were still right that it was not working. The key it grouped by was pinned to ADC0, so chain 1 never received a decision in practice. That is now fixed.

**On the authoritative source of DDC settings (your central question).** It is `DdcAssignment` plus the per-board codec, not `ReceiverManager`. For your xDMA work, the four files worth reading in order are `src/core/DdcAssignment.h`, `src/core/SliceStreamAllocator.h`, `src/core/codec/CodecContext.h`, and the codec for your board, `src/core/codec/P2CodecOrionMkII.cpp`, which the ANAN-G2 now inherits from.

`ReceiverManager` is still live for I/Q routing and receiver lifecycle, but it stopped being the DDC-settings authority. You are right that `ReceiverConfig::isPureSignalFeedback` and `isDiversity` are never written and never read. They are dead state from the earlier design and you should ignore them. `ReceiverManager::setDiversityEnabled` likewise has no production caller. Treating those fields as authoritative would have misled you, which is exactly what happened.

**On `applyDdcAssignment` not setting DDCs properly.** You were right when you wrote it. On 2026-06-04 that function had zero production callers; the wiring landed on 2026-07-24. It now sets rate, enable and sync correctly. It did not set the ADC field correctly until this week.

**On diversity, you are right and we have not implemented it.** The DDC0/DDC1 pair and its sync and enable bits are correct and Thetis-faithful, but the constraints you list are not: the DDC-to-ADC split, equal filtering across both chains, allocating diversity receivers first, and supporting more than one diversity receiver. There is a related gap on the DSP side, where WDSP's `pdiv[]` is a two-slot array keyed by a diversity id rather than the channel id and dereferences unallocated state above index 1. A proper allocator does not exist yet. Your n and n+1 with n even, n synchronous on ADC0 and n+1 on ADC1, filters identical, is the specification we will build against.

**What your comments found, now fixed:**

- Antenna-driven ADC routing existed only in the OrionMkII codec. `P2CodecSaturn` overrode `applyDdcAssignment` and never received it, and every test for the feature constructed the OrionMkII codec, so they passed while the feature was absent on the ANAN-G2. Saturn's override is deleted and it now inherits the working implementation, with an anti-drift test sweeping both codecs for agreement.
- `CodecContext::adcCtrl` was never seeded on Protocol 2, so it stayed 0 and the diversity pair sat on ADC0 twice. Thetis defaults `rx_adc_ctrl1` to 4, meaning DDC1 on ADC1 (`console.cs:15099`, encoding at `setup.cs:16934`). Now seeded and gated on the board's ADC count.
- The Alex filter analysis now follows the DDC to ADC1, so each chain gets an independent decision.
- Related discovery: nothing recomputed the DDC assignment when the antenna changed, so the routing was invisible even where it existed.

One correction worth recording: ADC count and filter-chain count are not the same thing. The ANAN-100D and 200D report two ADCs but are absent from Thetis's `setAlex2HPF` model list, so they have two ADCs and one driven filter chain. That needed a separate capability rather than reusing the ADC count.

**On a sample-rate budget.** Not implemented, and the current limits are structural only: DDC count and a per-SKU maximum. Your framing is better than what we have, because it captures transport bandwidth and WDSP headroom, which the structural limit ignores entirely. Worth its own issue rather than being buried here.

J.J. Boyd ~ KG4VCF

CT1IQI:
Thanks for comments; Will raise separate 'Issue' regarding total CPU and bandwidth resources budget from which DDCs and sample rates could be be attributed, not to forget about it. Regarding diversity reception disagree that WDSP poses a problem. WDSP, now available in version 2, facilitates a diversity mixer but there is no need to actually use it. E.g. in piHPSDR the diversity mixing is done on basis of the IQ signals from the two paired DDC/Filter/ADC chains in the app's own code. The resulting single IQ signal is fed to WDSP as if it were a simple receiving channel. There really is not much to the mixing, something like:
iVal = iVal0 + (div_cos * iVal1 - div_sin * qVal1);
qVal = qVal0 + (div_sin * iVal1 + div_cos * qVal1);
Whether wdsp does this internally, or Nereus does it, no difference really apart from more flexibility when you let Nereus do it.
It means basically when defining a diversity receiver attributing to it an amplitude and phase setting value, or the resulting sine and cosine settings.

CT1IQI:
And, i.m.h.o. diversity is a must have. It is a major benefit of having a SDR with dual antenna/adc/filter/ddc chains, like the G2.

boydsoftprez:
Out of draft and ready for a first review. This has never been reviewed, so a few pointers on where the risk actually sits.

The diff is large but it is one thing: the RX data plane moved from a single implicit receiver to N slices bound many-to-one onto DDC streams. The places most worth your attention:

- `SliceStreamAllocator` and `RadioModel::computeDdcAssignment`, plus the per-board codecs under `src/core/codec/`. This decides which slice lands on which DDC and ADC, and it is where three defects were already found: a codec subclass that silently overrode the routing, an unseeded `adcCtrl` that put a diversity pair on one physical input twice, and a filter analysis pinned to ADC0.
- `MasterMixer` and `AudioEngine` admission and withdrawal. Audio-thread ordering, no mutex in the callback, and a deliberate absence of timeouts.
- `TxSliceArbiter` and the TX authority paths, where mode, filter and antenna each resolve from a different source.

One class of bug recurred five times here and is worth grepping for specifically: a check that reads a different thing than the code decides on. A test asserting a field that stopped being the field that decides, a guard implemented and documented as wired but never connected, state written by nobody and read by nobody, and shims that round-trip through a variable while reaching no DSP. Round-trip assertions pass for all of them.

Known and deliberately unfixed, so not worth reporting: the per-slice antenna versus ADC routing defect written up in `docs/architecture/2026-07-30-per-slice-antenna-adc-routing.md`. It ships as a documented limitation and the fix runs as its own branch.

Bench verification has not been run. The merge is gated on that, not on this review.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
I agree and will certainly have diversity in Nereus “eventually”.


J.J. Boyd

boydsoftprez:
All three findings verified against upstream and fixed. Each was real, and the second and third were more interesting than they first looked.

**P1, synchronized DDC1 in the enable mask** (`71151176`). Confirmed. Both the PureSignal and diversity branches did `ddcEnable |= 0x03`. Measured before the fix: `0x7` on the PS path where upstream writes `0x5`, and `0x3` on the diversity path where upstream writes `0x1`. The Thetis citations quoted directly above each branch already said so, `DDCEnable = DDC0 + DDC2; SyncEnable = DDC1;` at console.cs:8267-8268 and `DDCEnable = DDC0; SyncEnable = DDC1;` at console.cs:8290-8291. `nDdc` deliberately left alone: on Protocol 2 it feeds one log line, and the pair does physically occupy two DDCs whatever the mask says.

**P1, Hermes PS branch keeping `streamDdc[0]`** (`904e675f`). Confirmed. The mapping is now set inside the plain-RX and diversity branches where it is true, and left at the idle sentinel in the PureSignal branch. New test file, since there was no Hermes-class codec test at all; four cases, because the three that must keep working matter as much as the one that was broken.

**P2, `psPaused` never written in production** (`dfec4d97`). Confirmed: test-only callers, while `PanadapterApplet` reads it to drive the PS HOLD pill. Driven from the assignment as you suggested rather than from the PureSignal coordinator, which is what the property's own declaration proposed. The codec decides whether a slice still has a DDC, so reading its answer cannot drift from it; a separate signal out of the coordinator would be a second opinion free to disagree. Coverage discriminates rather than just round-tripping: both slices pause when Hermes PureSignal keys, both clear on unkey, and neither pauses on an ORION-class radio where PS keeps every receiver.

**The part worth reading** (`4d8e5846`). Fixing the two P1s broke eight assertions across four tests, and every one of them was defending the defect.

Six confused the enable mask with the sync mask. In each case the Thetis citation sat directly above the assertion with the right answer in it, but a human-written summary line above the citation read "DDC0+1+2+3 enabled" or "DDC0+1 enabled+synced", and the assertion followed the summary instead of the upstream two lines below. One of them asserted DDC1 in the enable mask on the line after a comment saying "DDC1 is synchronized to DDC0" and the line before an assertion putting it in the sync mask.

Two confused DDC0 being enabled with the user stream owning it, including one named `hermes_class_p2_ps_mox_keeps_stream_zero_on_ddc0`, whose comment reasoned from "DDCEnable = DDC0 is unconditional on this family" to "so stream 0 does not migrate". The enable mask cannot answer that. GetDDC can: `tot = nME + (nDE << 1) + (nPSE << 2)` at console.cs:8560, so `tot == 5` is MOX on, diversity off, PureSignal on, and the Hermes-class case 5 body is empty at console.cs:8635-8636, leaving `rx1` and `rx2` at the -1 they were initialised to. RX1 is genuinely gone. That test is renamed and now asserts `psFwdDdc == 0` alongside, so the reason DDC0 is enabled is visible in the test itself.

Every corrected assertion records what it used to read and why that was wrong, because this was not a typo. A citation and an assertion disagreed, the assertion won, and the suite defended the defect long enough that correcting it broke eight tests at once.

**Behaviour change for the bench, not just a mask correction.** On Hermes-class radios including the ANAN-G2E, MOX with PureSignal now suspends every user stream including Slice A, not only the extra ones. Both pans grey and show PS HOLD during PS transmit. That is what Thetis does and what this code's own comments already described, but it will look different on air and is worth watching for.

Suite is 570/570 through the `all_tests` target, build exit code checked before running it.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Second round: all three verified against the code and fixed.

**P1, layout shrink orphaning slices** (`3a1fe193`). Confirmed. The loop only ever adds, so with `existing > target` its body never runs and slices whose `panKey` named a deleted pane were left pointing at nothing while still holding a DDC, a stream and audio.

Rehomed onto the first surviving pan rather than removed. A slice carries the operator's frequency, mode, filter and DSP state, and discarding that as a side effect of picking a smaller layout is destructive and was never asked for. An empty pan list is a no-op, because leaving a stale pan id beats pointing at an empty string no pane will match.

The reconciliation lives on `RadioModel`, not in the `MainWindow` lambda that has the defect. MainWindow is not constructible in the test harness, so logic put there cannot be tested at all, and two fixes have already shipped green and failed on the bench for exactly that reason on this branch. The lambda now calls one testable method.

AetherSDR was checked for a precedent and does not supply one: its `removePanadapter` sends `display pan remove` and lets the radio decide what becomes of the slices. NereusSDR owns slice and DDC allocation locally, so the workflow shape ports and the mechanics do not.

**P1, wideband on a shared chain** (`757b13b7`). Confirmed and fixed as an OR across every live slice on the chain. The incoming boolean is now ignored on purpose: `SliceModel` has already committed it before the signal fires, so the recompute reads the new value along with everyone else's and there is no second copy of the edge to drift.

**P2, capability gate** (same commit). Confirmed, and `widebandAdcs` turned out to be declared across the whole SKU table and read by nothing at all, so this is where it finally gets consulted.

One deliberate narrowing to flag. The gate tests `widebandAdcs <= 0` rather than `chainIdx < widebandAdcs`. A chain index is not an ADC index: ANAN-100D and 200D report two ADCs behind a single preselector chain, and comparing one against the other is the ADC-count-versus-chain-count confusion that has already produced defects on this branch. The zero test is unambiguous and covers the reported case; narrowing further needs the chain-to-ADC mapping settled first.

That surfaced something adjacent which is **not** fixed here and is recorded in the code: `setWidebandEnabled` takes an ADC index (`P2RadioConnection.h:335`) and this call site passes a chain index. They coincide on the 2-ADC 2-chain boards and diverge on ANAN-100D and 200D. Same root question, same follow-up.

The gate also treats "no board yet" as "cannot do wideband". That is correct rather than convenient: before connect there is no preselector to bypass and nothing to stream, and it does not get stuck, because `SpectrumWidget::recomputeExtendedMode` runs on sample rate and bandwidth changes, so connecting re-evaluates with the board known.

Every fix was verified red first by reverting or stubbing the production change and confirming only the behavioural cases failed while the controls stayed green. Suite is 572/572 through the `all_tests` target, with the build's exit code checked before it.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Round three: both confirmed and fixed in `d14bbbd4`. Both were holes in the previous round's own fixes rather than pre-existing defects, which is the more useful finding of the two.

**P1, the layout fix reached one of three call sites.** Correct, and worse than the comment says: there are three places that apply a layout, not two. Session restore, the View menu, and the `+PAN` dropdown, each with its own copy of the template table, the id list and the add-only loop. Round two's fix went into the View-menu copy and no grep was run for the others, so the defect stayed fully live through the entry point operators actually use.

Fixed by centralizing rather than by copying the fix a second time. `applyPanLayout` now owns apply-then-reconcile for both interactive paths, and `panIdsForLayout` owns the template table that had three copies, so a later change cannot land in one path and miss two. Session restore shares the table and deliberately nothing else: it runs at startup before a radio is connected and before any slice exists, so there is nothing to rehome and the add-loop would manufacture a slice pre-connect.

**P2, wideband recomputed on the wrong trigger.** Correct. The handler runs on `widebandExtensionRequested` edges, and removing the sole requester changes the chain's answer with no property moving at all. The push is now `pushWidebandStateForChain`, called from the property handler and from `removeSlice`, which captures the chain before the slice leaves `m_slices` because afterwards nothing can say which chain to recompute. Two tests: removing the only requester clears the chain, removing a slice another still needs does not. Verified red with only the removal case failing.

Worth stating plainly, since it is the pattern rather than the bug: round two's layout fix was applied to the call site named in the review and nowhere else. Grepping for siblings is cheap and would have caught it. Centralizing was the right answer the first time.

Suite is 572/572 through the `all_tests` target, with the build's exit code checked before it rather than inferred from the ctest summary.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Round four: both confirmed and fixed in `a3da778e`. Both were gaps in the previous rounds' own fixes, which is the finding worth acting on, so this closes the class rather than the two instances.

**P1, chain migration.** Correct. The chain's wideband answer changes on more inputs than any one signal names: a slice's request moving (round 2), a slice being removed (round 3), and a slice migrating chains when its antenna moves its DDC to the other ADC (round 4). Each was wired as its own hook, and each round found the next one missing. That is not a bug pattern, it is a design one.

`reconcileWidebandForAllChains` now recomputes every chain from current state, called from `publishDdcAssignment`, which already rebuilds DDC, chain and psPaused the same way. Migration is covered by construction rather than by remembering it exists.

The round 3 hook in `removeSlice` is **deleted**, not left alongside. Removing it and watching all nine wideband tests stay green is what proves the sweep subsumes it; keeping both would be two owners for one fact, which is the shape this branch keeps paying for. I checked the path is a direct call chain on the same thread (`requestDdcAssignment` to `invokeCodecDdcAssignment` to `publishDdcAssignment`) before deleting, so there is no window where the sweep has not yet run.

**P2, empty pans.** Correct, and it was a regression my own round 2 rehoming created rather than a pre-existing defect. `pansWithoutSlices` asks occupancy directly, counting a pan once however many slices share it. Shrink now runs before grow so the occupancy question sees settled state. I also dropped the duplicated `maxSlices` arithmetic, since `addSliceOnPan` already enforces the cap and emits `sliceAddRejected` with an operator-facing reason, and restating it was a second copy of that policy.

Four new tests, including the migration case verified red with a precise signature: `chainIndex` and `widebandActiveForChain` were both already correct, and only the wire push for the vacated chain was missing.

Ledger across four rounds, since it is the useful part: eight findings, four of them introduced by fixes for earlier findings. Rounds 3 and 4 found nothing pre-existing, only my own gaps. That is the argument for making this a sweep, and also the reason I am not assuming round five is clean.

Suite is 572/572 through the `all_tests` target, build exit code checked before it, no unused-variable warnings.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
## Round 5 addressed (`39f994be`)

Both findings were P2. One of them changed shape once I chased it properly.

### P2: Protocol 1 boards reached extended mode

My first fix was a gate in `RadioModel::widebandActiveForChain`, rejecting the connection when it is not a `P2RadioConnection`. It worked, and it broke `tst_alex_bpf_policy_push` and `tst_pan_wide_badge`, whose `BpfMockConnection` derives from plain `RadioConnection`. Two tests failing on a fix that behaves correctly is the tell that the gate was in the wrong place.

`BoardCapabilities` already carries a `.protocol` field. Both rows involved declare Protocol 1 and advertise wideband ADCs anyway:

```
kAngelia  (ANAN-100D): .protocol = Protocol1,  .widebandAdcs = 2
kOrion    (ANAN-200D): .protocol = Protocol1,  .widebandAdcs = 2
```

Every other Protocol 1 row sets `0`, with the comment "wideband mechanism differs; deferred to 3F-W". These two contradicted their own protocol field. NereusSDR has no P1 wideband receive path, so the value let extended view bypass the Alex preselector for a stream that never arrives: receive filtering lost, nothing gained.

The fix is the table. Both rows go to `0`, and the connection cast is gone. The two mock-based tests are untouched, because there is nothing left to break them.

Tested as an invariant rather than per SKU:

```
no_protocol1_board_advertises_wideband_adcs()
```

sweeps `BoardCapsTable::all()`, so a SKU added later is covered without anyone remembering to extend a list. Verified it bites by stashing only the table fix: it fails naming ANAN-100D and its value. Had it existed, this would have been caught at authoring time rather than at review.

### P2: expanding a layout created slices it did not need

After a shrink every slice is co-hosted on `pan-0`, so the slices the empty pans want already exist. `addSliceOnPan` was spending the `maxSlices` budget making new ones: on a five-slice radio that fills `pan-1`, hits the cap, and leaves `pan-2` and `pan-3` empty with a surplus fifth slice in the model.

`RadioModel::spreadSlicesOntoEmptyPans` moves the surplus first; `applyPanLayout` creates only for what is still empty afterwards. A pan holding exactly one slice is never raided, or expanding would relocate the hole rather than fill it. Occupancy is recounted each iteration, because the previous move changed the state the next decision rests on.

Verified red by stubbing the spread to a no-op.

### Evidence

Full suite `572/572`, build exit `0`, built via the `all_tests` target rather than a bare `ctest` so nothing ran stale.

### Note on this round

Both findings sit in multi-slice paths that no operator has run yet. Round 5 is the first round where the root cause turned out to be a data-entry error in a capability table rather than a logic defect, which is a reasonable place to hand the rest to a radio. Bench verification is next.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Re-triggering against current HEAD. `74f72b80` landed after the last request and is documentation only: bench matrix rows 18-25 covering what the suite cannot reach. No source change since `39f994be`.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
## Rounds 6 and 7 addressed (`dab74491`, `c4ef8250`)

Five findings. Also a bench report that landed alongside them and turned out to share the shape.

### Bench: the bottom bar was being blanked, and not only by Phase 3F

Pressing TUNE with PureSignal active replaced the entire bottom bar with one line of text for six seconds. The message was right and the surface was wrong: `QStatusBar::showMessage` hides every non-permanent widget while a message is up, and `buildStatusBar` adds the whole bar as one such widget. So every notice in the tree cost the operator the CH pill, the PureSignal indicator, the radio name, CAT/TCI state, the PA and TX badges and the clock, and the two that fire during transmit cost it mid-transmit.

That is **all twelve** `showMessage` call sites, not just the Phase 3F ones. All twelve moved to a new `StatusToast`. The suspended-streams notice additionally takes itself down when the streams return rather than aging out after the operator has already unkeyed. Nothing calls `showMessage` now, and `buildStatusBar` says why.

### P1: the untagged IQ tap forwarded every stream

Since Sub-Epic I the fork fired once per stream, while its only subscriber still carried its original precondition in a comment:

```
// Phase 18: RadioModel::rawIqData fires only for RX1 (slice 0).
constexpr int kReceiver = 0;      // TciServer onRawIqDataReceived
```

Sub-Epic I made that false. A client on `iq_start:0` was fed consecutive blocks from different DDCs on different bands under one receiver header. Silent, because every frame is individually well-formed. The fork is now a named method publishing the untagged tap for stream zero only.

### P2: wideband and the live connection protocol

Round 5's table correction was necessary and not sufficient, exactly as you said. A row's `.protocol` is what the board usually speaks, not what this connection speaks. `ANVELINAPRO3` and `REDPITAYA` have real Protocol 1 codecs and both resolve to `kOrionMKII`, which declares Protocol2 with `widebandAdcs = 2`.

Gated on `m_lastRadioInfo.protocol`, deliberately not the `qobject_cast` I reached for first. That version was untestable against the `RadioConnection`-derived mocks, which is how the original row error survived, and it broke both suites. Reading the `RadioInfo` costs nothing and both suites pass with no cast.

### P2 x2: TCI was outside the surface it had negotiated

`tx_frequency` came from receiver 0 under a comment saying the caller would decide once 3F landed the real TXFreq logic. It landed, and the test was wrong in both directions: tuning Slice A advertised A as the transmit frequency while the transmitter was on B, and tuning B advertised nothing. Now sourced from `TxSliceArbiter`, with a handoff wire so a TX move republishes without waiting for a retune that may never come on a parked slice.

`trx_count` clamped rather than raised. The count is the locked half of the decision per design doc §1.2, and raising it means widening the per-receiver init burst, which the doc puts out of scope. The `2` was a bare literal nothing else could read, which is precisely why the two drifted; it is now `kExposedReceiverCount` and both sides read it.

The two interact, and a fix for either alone breaks the other, because `tx_frequency` carries no receiver index and `vfo:N` does. The untagged pair is now separately callable so a TX-bound internal slice publishes its transmit frequency while staying silent on everything tagged.

### P2: the restored sample rate never reached the DDC

`restoreFromSettings` called `setSampleRateHz`, a plain property setter. Applied after the restore rather than inside it, because the restore sets frequency first and completes the allocator rebind, so the transaction has to run against the stream the slice ended up on. Both restore paths, launch and band change.

### Two existing cases were asserting the defect

The `trx_count` clamp broke two cases that required frames for receiver 2. Worth being explicit that neither was relaxed.

Both pin stable slice ids: a survivor broadcasts its own id, never its new list position. Correct and worth keeping. They demonstrated it at id 2, which meant asserting the server would tell a client about a receiver it had not allocated and could never address back, since inbound resolves through `sliceById(N)` per the mapping decision documented in `TciProtocol.cpp`.

I checked whether the wire index could be decoupled from the slice id instead, which would have made both true at once. It cannot: that mapping is protocol-facing and every inbound per-rx handler depends on it. So both retarget to id 1, where removing slice A leaves B at list position 0 with id 1 and separates id from position exactly as before. Both gained an assertion that id 2 stays off the wire, so the clamp is pinned rather than merely uncontradicted.

### One thing worth flagging

Four separate test files hand-roll "declare a radio", and each declares a different subset of what a radio is. Round 5 added the board to two of them; round 6 found the protocol missing from those two plus two more. A shared helper would end it, but that is test infrastructure beyond these findings, so it is noted rather than built.

### Evidence

Full suite `576/576`, build exit `0`, via the `all_tests` target. Every new case verified red first, and the three that failed on the first full run are described above rather than quietly fixed.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Connecting produced no waterfall and no audio until the operator nudged the
VFO. Reported from a live Hermes Lite 2 over a routed link, and reproduced
on an ANAN-G2E, so it was never P1-only.

## Symptom

    P1: first ep6 frame received from "::ffff:x.x.x.17" (1032 bytes)
    P1: first iqDataReceived emit; rx= 0 samples= 76
    ReceiverManager: first feedIqData dropped; hwReceiverIndex= 0 map= "(empty)"

`map= "(empty)"` is the failure. An empty `m_hwToLogical` means no receiver
was active, and `rebuildHardwareMapping` only maps active ones.

## Root cause

`computeDdcAssignment` returned a default `DdcAssignment` when no codec had
been selected yet, and a default `DdcAssignment` has every `streamDdc` at -1.

That -1 is not a spare value. It already means something specific and
load-bearing: the codec was asked, and answered that the radio has stopped
streaming that DDC. `publishDdcAssignment` is required to act on it, and
5851998a exists because it once did not, so PureSignal feedback reached a
panadapter and the speakers through a receiver left active.

So "nobody has been asked yet" and "the radio stopped every stream" arrived
at the decision point indistinguishable. `connectToRadio` activates receiver
0 (:5527) and then binds the slice pool (:5579), and that bind's
`requestDdcAssignment` tail runs before the connection object exists (:5855)
and before `wireConnectionSignals` installs the codec (:7674). Every connect
therefore published a fabricated suspension and deactivated the receiver it
had activated forty lines earlier. Nothing re-ran the assignment once the
codec arrived, so the mapping stayed empty until the operator moved the VFO
and `frequencyChanged` drove a bind that finally had a codec to ask.

## Fix

Two halves, both load-bearing.

1. `computeDdcAssignment` returns `std::optional`, with `std::nullopt` for
   "no codec has been asked". `invokeCodecDdcAssignment` declines to publish
   that, leaving the receiver `connectToRadio` seeded intact. The type now
   forces callers to decide, rather than letting one sentinel carry two
   facts.
2. New `ReceiverManager::ddcCodecChanged`, wired to `requestDdcAssignment`,
   so the codec arriving re-runs the assignment and its real answer lands
   before the first frame rather than at the first VFO tick. Wired on the
   setter because four installers go through it, a `codecChanged` lambda
   plus a catch-up poll per protocol.

`reconcileWidebandForAllChains` still runs on the no-codec path. It reads no
assignment, recomputing each chain from the live slices, and
`requestDdcAssignment` is also the coalescing point for slice removal, so
skipping it left the radio streaming a wideband chain whose last requester
was gone.

The suspension gate itself is untouched. A codec that answers -1 still
deactivates.

`publishDdcAssignment` now logs both inputs to the activation decision per
stream, not just the outcome. The two conditions fail for unrelated reasons
and produce an identical symptom, which is what sent the first pass at this
after the wrong one.

## Second symptom, same cause

The fabricated assignment also satisfied the suspension block, so every
connect emitted `streamsSuspended` and raised a six-second warning toast
reading "Slice A has no receiver: this radio has no DDC free for them right
now." Covered by its own test case, since it is driven from a different
block and a change could restore one without the other.

## Tests

`tests/tst_connect_routes_first_iq.cpp`, five cases: P1/HL2 routes the first
frame, P2/Saturn likewise, empty streams stay deactivated through connect, no
spurious suspended-stream notice, and a real codec -1 still deactivates.

Each was confirmed to fail before the fix. The toast case was additionally
confirmed against a tree with only the `std::nullopt` half reverted, which is
what shows the two halves are independent.

`tst_suspended_stream_has_no_receiver` stays green. Full suite 582/582.

## Bench verification

Confirmed on hardware, both protocols. The line that previously read
`dropped; map="(empty)"` now reads:

    ReceiverManager: P1 codec set
    publishDdcAssignment: stream 0 slices=1 streamDdc=0 -> activate
    P1: first ep6 frame received from "::ffff:x.x.x.17" (1032 bytes)
    ReceiverManager: first feedIqData forwarded; hw= 0 logical= 0 wdspChannel= 0 samples= 76

ANAN-G2E on Protocol 2 shows the same shape, `first feedIqData forwarded;
hw= 0 logical= 0 samples= 476`. Waterfall and audio are live at connect with
no retune.

## Not addressed here

Two things the restored log surfaced, both pre-existing and neither caused by
this change:

- `first feedIqData dropped; hwReceiverIndex= 1 map= "hw0->rx0"` on HL2. The
  radio is told `activeRx=4`, from the HL2 codec's unconditional
  `p1RxCount = 4` ("RX4 used for puresignal feedback"), which is
  mi0bot-faithful. Slots 1 to 3 were always dropped. This is newly visible
  only because the one-shot first-drop log slot was previously consumed by
  the hw0 failure.
- On the ANAN-G2E the stream pool opens 4 streams while only 2 receiver
  objects are created, because the receiver cap comes from discovery and the
  pool comes from the capability row. Inert with one slice, but
  `activateReceiver` returns silently for a receiver that does not exist, so
  it would reproduce this same empty-map failure once multi-slice binds a
  slice to stream 2 or 3. Tracked separately.

J.J. Boyd ~ KG4VCF

PR #305, opened 2026-07-26
build: exclude test executables from the default build target

boydsoftprez:
Every `tst_*` executable was part of the default `all` target, so any source change touching `NereusSDRObjs` relinked all 513 of them. Each test statically links the whole application (~22 MB of `__TEXT`), so verifying a one-line edit cost roughly 32 minutes of linking.

Measured on this branch, after touching `src/core/AppSettings.cpp`:

| | Before | After |
|---|---|---|
| `cmake --build build` | ~32 min | **1.64 s** |
| Test binaries linked | 513 | **0** |

Ported by hand from `6ed89682` on `feature/rfkit-rf2ks-applet` rather than cherry-picked, because that commit bundles three unrelated changes and only one applies here:

- **`EXCLUDE_FROM_ALL` + `all_tests` aggregate** — taken.
- **LTO default ON→OFF** — skipped. `main` has no `NEREUSSDR_ENABLE_LTO` option at all; it's added by an earlier commit on that branch. Every `CMakeLists.txt` hunk in `6ed89682` is LTO-related.
- **Partial-region Metal overlay upload** (`SpectrumWidget.cpp`) — skipped. Unrelated GPU perf work that belongs with its own branch and wants bench verification.

The `add_executable` line needed adapting: `main` now passes `$<TARGET_OBJECTS:nereus_test_sandbox>` where that branch still lists `TestSandboxInit.cpp`.

Also includes the CI fix from `1d3650cc`, which is **not optional**. Without it this change reintroduces the exact P1 raised in review on #291: ctest finds the registered test definitions but no binaries, reports every one as "Not Run", and **exits 0** — silently green CI.

The Linux gate is correct because `ci.yml:458` is the only place `-DNEREUS_BUILD_TESTS=ON` is passed, so `all_tests` doesn't exist on the macOS or Windows jobs. I reworded the comment to say that, rather than the original claim that those jobs configure tests OFF.

Verified: `cmake --build build --target all_tests` then `ctest` gives 513/513 passed, 0 failed, nothing reported Not Run.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -493,6 +493,23 @@ jobs:
         shell: bash
         run: cmake --build build -j$NUMBER_OF_PROCESSORS
 
+      # Build the test executables explicitly.  Every tst_* executable is
+      # now EXCLUDE_FROM_ALL, so a routine `cmake --build build` builds
+      # only the main app (seconds, not the ~32 minutes it takes to link
+      # 513 test binaries).  CI must therefore ask for the `all_tests`
+      # aggregate before invoking ctest, or ctest finds the registered
+      # test definitions but no binaries and reports every one as
+      # "Not Run" -- which exits 0 and would silently pass.
+      #
+      # Gated on Linux to match the configure step: only the Linux job
+      # passes -DNEREUS_BUILD_TESTS=ON (ci.yml:458), and `all_tests` is
+      # created by tests/CMakeLists.txt, which is only add_subdirectory'd
+      # when that option is ON.  Asking for the target elsewhere would
+      # fail with "No rule to make target".
+      - name: Build test executables (Linux)
+        if: runner.os == 'Linux'
+        run: cmake --build build -j$(nproc) --target all_tests
+
       # ------------------------------------------------------------------
       # Test
       # ------------------------------------------------------------------
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,8 +44,13 @@ that matches our conventions.
    unit-test suite, read
    [docs/development/fast-test-loop.md](docs/development/fast-test-loop.md)
    first: every test executable statically links the whole application, so
-   building all of them costs about 32 minutes. Build the one target you
-   care about, or use `ctest -L <subsystem>`.
+   building all of them costs about 32 minutes. Test binaries are
+   `EXCLUDE_FROM_ALL`, so always build a target before running `ctest` --
+   otherwise you are testing stale binaries:
+
+   ```bash
+   cmake --build build --target tests_core && ctest --test-dir build -L core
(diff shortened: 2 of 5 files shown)


boydsoftprez:
Both Codex findings addressed. Details, including where the evidence is weaker than I'd like.

### 1. `tests/CMakeLists.txt` — labeled tests could run stale binaries (this PR)

Confirmed, and it was a real defect in the workflow this epic documents. Reproduced before fixing by recording `tst_smoke`'s mtime across a source edit:

| workflow | `tst_smoke` mtime | outcome |
|---|---|---|
| `cmake --build build` then `ctest -L …` | unchanged | **stale binary, false green** |
| `--target tests_unclassified` then `ctest -L …` | updated | fresh |

So a developer could edit source, build, run the subset, and get a green that tested their *previous* code.

Fixed in 46cd9950 by emitting one build target per derived label — `tests_core`, `tests_models`, `tests_gui`, `tests_unclassified` — from the same configure-time derivation that produces the ctest labels, so targets and labels cannot drift apart. Verified a subset target builds only its subset: after a `src/core` edit, `--target tests_unclassified` relinks exactly 5 binaries, not 513.

`docs/development/fast-test-loop.md` and `CONTRIBUTING.md` now document the safe form:

```
cmake --build build --target tests_core && ctest --test-dir build -L core
```

Also removed the guide's "Known wart" section, which claimed `EXCLUDE_FROM_ALL` was absent and pointed at three unmerged branches. The first commit on this PR made that untrue.

### 2. Compressed watchdog margin (carried over from #304)

Confirmed correct on the numbers: the compressed silence threshold was 20 ms, `kWatchdogTickMs` is 25 ms, and `P1FakeRadio` streams every 10 ms (`P1FakeRadio.cpp:54`). A threshold below the detector's own sampling period has no margin by construction.

Raised to 150 ms / 300 ms in f2b90d69, leaving 6 watchdog ticks and 15 stream intervals of headroom. Cost is 1.28 s → 4.58 s for that test, still well below the 53.5 s it started at, and it no longer sets the suite's parallel floor.

**Weaker evidence here, stated plainly:** I could not reproduce a failure with the old 20/50 values on this machine — 0 failures in 24 runs at 8-way parallelism under 12 spin loops, identical to the new values. So this is not a fix for an observed flake. It is a fix for a threshold that is wrong by construction, on the reasoning that CI runners are slower and more contended than an M-series laptop. If that reasoning is wrong, the change costs 3 seconds of test time and nothing else.

### Note

The review was against `be211e72`, the first of what are now three commits. `f2b90d69` and `46cd9950` have not been reviewed yet.

J.J. Boyd ~ KG4VCF

boydsoftprez:
@codex review

PR #306, opened 2026-07-27
fix(p2,discovery): stop wedging the ANAN-G2E on disconnect

boydsoftprez:
## Problem

Disconnecting from an ANAN-G2E (HermesC10, fw 110, Protocol 2) left the radio
unreachable at layer 2. Not just discovery: the MikroTik on the same segment
could not ARP or ping it either, and the switch port showed no traffic at all.
Only a physical power cycle recovered it. Thetis on the same radio and the same
network never triggered it. Not reproducible on HL2, ANAN-G2 (Saturn), or
Anvelina PRO3.

## Root cause

Our disconnect was fine. The stop frame is byte-identical to Thetis's
(`byte4=00`, frequencies preserved, seq 0) and the radio acts on it, stopping
its stream cleanly between bursts.

What differed is what came next. Teardown reopens the ConnectionPanel, whose
constructor auto-scans, so a broadcast discovery burst hit the radio 7-15 ms
after `run=0`. Thetis goes completely silent after its stop frame. Both
observed lockups landed inside that window: one radio died before answering the
first burst, the other answered burst 1 and was dead before burst 2 ~850 ms
later. Probing a radio while its stop-transition state machines are still
settling is a race Thetis never enters.

## Changes

**1. Post-disconnect discovery quiet period.** `RadioDiscovery::holdOffScans()`
defers, never drops, any `startDiscovery()` or `probeAddress()` issued within
3 s of a teardown. The panel still populates and Connect flows still probe,
just after the radio has finished stopping. Covers every teardown path: user
disconnect, failed-connect watchdog, and quit.

**2. P1 discovery probe no longer impersonates P2 commands.** The probe is
broadcast to UDP 1024, which is also the P2 General command port. P2 gateware
claims any port-1024 datagram whose byte 4 is zero and parses the rest as
config, so our all-zero tail was clearing `HW_timer_enable` (the board's ~2 s
deadman), `PA_enable`, and `Alex_enable` on every P2 radio on the subnet, every
scan. Byte 4 is now `0xFF`, which no P2 command claims. Deliberately not `0x02`
(the P2 discovery command, doubles every radio's reply load) and not `0x03`
(reaches the EEPROM set-IP path on a broadcast frame). Verified no P1 gateware
reads byte 4: Hermes v3.3, Angelia, Orion, ANAN-10E/100B and HermesC10 all
match only the command byte in `Rx_MAC.v`. Confirmed on hardware after the
change: HL2 still discovered over P1, Saturn reply count back to normal.

**3. Stop polling the radio during RX.** Wire capture showed our command
cadence at 27-280x Thetis's. Thetis never polls; only `CmdGeneral` is periodic,
from `KeepAliveLoop()`, to feed the deadman. Our 100 ms heartbeat wheel was
purely additive, since every state change is already pushed immediately. The
wheel is gated on MOX rather than removed, preserving the 3M-1a bench finding
about TX carrier, which is a transmit concern only.

**4. `writeDatagram` return is now checked.** It was discarded, so
`"SendStop complete"` was logged whether or not the frame reached the wire.
Three rounds of debugging treated that line as evidence. It caught
`wrote -1 "Unable to send a message"` on the first run and proved the radio was
ARP-dead rather than ignoring us.

## Verification

Two consecutive disconnect/reconnect cycles on the G2E, no power cycle:

| | cycle 1 | cycle 2 |
|---|---|---|
| session | ~7 s | 54 s |
| scan after `run=0` | +2.32 s | +2.33 s |
| reconnect | clean | clean |

Cycle 2 ran longer than the 48 s session that wedged the radio before. The old
build never survived two cycles in a row.

Tests: `tst_p2_regression_freeze`, `tst_p2_mox_wire`, `tst_p2_codec_orionmkii`,
`tst_radio_discovery_probe`, `tst_radio_discovery_parse`,
`tst_flex_radio_discovery_broadcaster` all pass. Two new regression cases cover
the probe pad (asserts byte 4 avoids `0x00` and the whole `0x02`-`0x06` command
range) and the quiet period (probe completes, but only after it expires).

## Known limits

The gateware race itself was never pinned. The full TX chain reads
orphan-tolerant: `udp_send`, `ip_send`, `mac_send`, and `arp.v` all
self-complete, and the discovery handshake has a timeout. The one structural
suspect that could not be cleared is the TX arbiter in `network.v`, which
commits without a timeout: once `tx_ready` latches, only wire activity releases
it. That matches "mute with link up" exactly, but no requester was found that
can hang it.

So this stops us tripping the fault; it does not fix the firmware. Another host
broadcasting discovery at the wrong instant could in principle still wedge a
stopping G2E. Full investigation, including everything eliminated and its
gateware citations, is in `captures/g2e-disconnect-NOTES.md`. A report to
N1GP/TAPR is worth sending separately once the mitigation has a day of normal
use behind it.

Two unrelated defects found while diffing and deliberately left for follow-up:
`CmdTx` byte 50 bit 5 (documented by Thetis as Saturn balanced-input) is set by
default on Hermes-class boards, and RX2/RX3 NCO phase words are sent as zero
where Thetis sends real values.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/captures/g2e-disconnect-NOTES.md
+++ b/captures/g2e-disconnect-NOTES.md
@@ -155,18 +155,89 @@ period and it is what feeds the board's 2 s deadman.
 
 ---
 
-## Status
+## Second change: the P1 discovery probe
 
-**Unproven.** Cadence is the last wire-level difference standing after the others
-were eliminated against the gateware, not a demonstrated cause. The mechanism by
-which excess C&C traffic would leave the board's Ethernet path dead is not
-established; `sdr_send.v:138` (`if (!run && state > SEND) state <= IDLE`) is the
-only escape from a stalled send state and ARP shares a single TX arbiter with UDP
-(`network.v:341-344, 416-421`), which would explain total layer-2 silence, but
-that remains a hypothesis.
+Found after the cadence work, and the stronger suspect of the two.
 
-Next bench run: connect, use, disconnect, then confirm the radio still answers
-ARP and accepts a reconnect without a power cycle.
+Our P1 probe is broadcast to UDP 1024, which is also the P2 General command
+port. `General_CC.v:90/106` claims any port-1024 datagram whose byte 4 is zero
+and parses the rest as a General command. Our probe is `EF FE 02` followed by 60
+zero bytes, so a P2 radio accepts it and latches the padding as config: byte 38
+clears `HW_timer_enable` (freezing the ~2 s deadman), bytes 58/59 clear
+`PA_enable` and `Alex_enable`. `RadioDiscovery.cpp:441-445` sends it to both the
+directed subnet broadcast and 255.255.255.255 on every scan, and the app log
+shows a scan immediately precedes every connect, so it fired **every time** —
+matching the 100%-reproducible symptom far better than cadence does.
+
+Fix: byte 4 = `0x02`, so `General_CC` bails at its command check. Verified that
+no P1 gateware reads byte 4 (Hermes v3.3, Angelia, Orion, ANAN-10E/100B,
+HermesC10 all match only the command byte in `Rx_MAC.v`).
+
+This one was invisible in every capture we took: the mirror filter was
+`filter-ip-address=192.168.109.198/32`, which cannot match a broadcast
+destination. It was found in the application log instead. **If you are hunting a
+discovery-related problem, do not trust an IP-filtered capture.**
+
+---
+
+## Status: NOT FIXED — but the failure window is now pinned
+
+Bench run 2026-07-27 with both changes in. One cycle survived, the next did not,
+so the bug is **not deterministic** as originally believed, and neither change
+is sufficient.
+
+| session | duration | I/Q packets | outcome |
+|---|---|---|---|
+| 16:38:44 → 16:38:54 | 10 s | 2087 | reconnect OK |
+| 16:39:03 → 16:39:51 | 48 s | 9568 | **locked up** |
+
+### The important new fact
+
+**The radio survives the disconnect and dies roughly one second later, while
+idle, during the post-disconnect scan.** It is not dying at the disconnect.
+
+```
+16:39:51.448  P2: Disconnected. I/Q packets: 9568
+16:39:51.455  Scanning NIC "feth3169"
+16:39:51.497  P2 response from .198  ANAN-G2E fw: 110    <- alive, replies x4
+16:39:52.325  P2 response from .45   (Saturn only)       <- G2E gone
+16:39:53.141  P2 response from .45   (Saturn only)       <- G2E gone
+```
+
+The surviving cycle looks identical through the first scan attempt (four G2E
+replies) and then keeps answering attempts 2 and 3. The failing one answers
+attempt 1 and is dead by attempt 2. So the kill happens in a sub-second window
+after `run=0`, with the radio otherwise idle and only discovery traffic on the
+wire.
+
+This retires every "the stop frame is malformed" theory for good: the stop frame
+is byte-correct, the radio acts on it, answers discovery afterwards, and *then*
+dies.
+
+### A regression this run introduced, since fixed
+
+The first version of the probe pad used byte 4 = `0x02`. Byte 4 is also the P2
+*command* byte in `sdr_receive.v`, and `2` is the discovery command, so every P1
+probe became a second discovery request. Verified in the app log: 2 replies per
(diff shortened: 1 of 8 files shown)


boydsoftprez:
Good catch, and the bench data confirms it was already costing us in the normal case, not just the pathological one.

The quiet period was armed on entry to `teardownConnection()`, but `run=0` does not leave until `teardownWorkerThreadedConnection()` dispatches `P2RadioConnection::disconnect()` onto the connection thread, after the audio / WDSP / TX shutdown. From the run that verified this PR:

```
17:31:38.149  P2: SendStop sent
17:31:40.473  Scanning NIC "feth3169"
```

2.32 s of the intended 3 s. 680 ms lost. And as you note, `teardownWorkerThreadedConnection` may block up to `kDispatchTimeoutMs` (3000 ms) on a connection thread stuck in its `onReadyRead` drain loop, which equals the entire holdoff and would let the deferred scan fire immediately after `run=0`.

Fixed in 2dfe7f65 by arming again after `teardownWorkerThreadedConnection()` returns. The entry arm is kept so nothing scans during teardown; the second arm guarantees a full quiet period measured from the stop frame. `holdOffScans()` keeps the later deadline, so arming twice only ever extends.

Added a regression test for the property that makes double-arming safe: a shorter `holdOffScans()` must not pull the deadline in, a longer one must push it out.

@codex review

boydsoftprez:
Round 2 addressed in 14a3496a. The MOX one is a good catch and was a genuine hazard I introduced.

**P1, MOX-off retransmission.** Confirmed, and the existing code comment is the proof. `setMox()` early-returns on an idempotent call, justified by:

> the 100 ms high-priority periodic cadence already re-emits the current m_mox state on every tick

Gating that cadence on `m_mox` removed exactly the safety net the early-return depended on. Unkey sends one MOX-off datagram; if lost, the heartbeat has already stopped, later `setMox(false)` calls early-return without sending, and the `CmdGeneral` keepalive feeds the firmware deadman so it will not self-clear. Radio stays keyed. Fixed with a bounded 1 s MOX-off grace window so the heartbeat retransmits the off state ~10 times instead of once. It can only be open right after a transmission, so RX-idle polling does not return.

**P2, shared holdoff.** Confirmed, `AddCustomRadioDialog.cpp:589` does construct its own instance. The deadline is now process-wide, which matches what it describes: a property of the radio and the wire, not of one object. Side effect worth noting, it now outlives a single test function too, which surfaced as `timeoutEmitsProbeFailed()` failing the moment the static landed; added a `clearHoldOffForTest()` hook under the existing `NEREUS_BUILD_TESTS` block, called from the suite `init()`.

**P1, quiet period timing.** This one was already fixed in 2dfe7f65, which is in the diff you reviewed. The comment points at `RadioModel.cpp:7512`, the arm on teardown *entry*, which is retained deliberately so nothing scans during teardown. The second arm is at line 7818, after `teardownWorkerThreadedConnection()` returns, and that is the one measuring the full window from the stop frame. `holdOffScans()` keeps the later deadline, so arming twice only extends. If you still see a gap there I would like the specific path.

Tests: 8/8 in `tst_radio_discovery_probe`, 6/6 across the P2 and discovery suites.

@codex review

boydsoftprez:
Fixed both open findings from the review of `14a3496a` in `40202690`.

**Monotonic MOX-off grace** (`P2RadioConnection.cpp:889`) and **monotonic discovery holdoff** (`RadioDiscovery.cpp:159`): both stored their deadline as `QDateTime::currentMSecsSinceEpoch()` arithmetic, so an NTP step moved the deadline underneath them. Both now use `QDeadlineTimer` with `Qt::PreciseTimer`, which measures against a monotonic source. Default-constructed `QDeadlineTimer` is already expired, so the `m_moxOffGraceUntilMs != 0` never-armed sentinel is gone.

Public API is unchanged: `holdOffScans(chrono::milliseconds)` and `holdOffRemainingMs()` keep their signatures and their extend-never-shorten semantics, so the existing tests needed no edits.

The three earlier findings were already addressed on this branch: the quiet period is re-armed after the protocol disconnect completes (`RadioModel.cpp:7804-7819`), the holdoff is process-wide via a static, and the MOX-off retransmission window is what `withinMoxOffGrace()` implements.

28/28 tests covering `P2RadioConnection` and `RadioDiscovery` pass, including `holdOffScansExtendsButNeverShortens` and `holdOffIsSharedAcrossInstances`.

@codex review

PR #309, opened 2026-07-28
fix(attribution): the author-tag checker never ran locally, and could invent violations

boydsoftprez:
The inline author-tag preservation check has four defects. It has never
run outside CI, and when it does run without every upstream clone
present it can report attribution violations that do not exist.

Found while trying to run it from a worktree during unrelated work.

## The checker never ran locally

`THETIS_DIR` defaulted to `REPO.parent.parent.parent / "Thetis"`, which
resolves to `/Thetis` from a normal checkout and `<repo>/Thetis` from a
linked worktree. Neither exists, so `main()` took its FATAL-and-return
branch every time. `scripts/git-hooks/pre-commit` had the same bug
implemented separately, probing a fixed `../`, `../../`, `../../../`
list and bailing before it ever invoked the script. CI was unaffected
throughout because it supplies both paths by env var.

Both now walk ancestors, so neither needs to know its own nesting depth.

## It could invent violations

`resolve_upstream` fell back from the mi0bot clone to the ramdor clone.
mi0bot/OpenHPSDR-Thetis is a fork of ramdor/Thetis, so both carry files
with the same names whose contents and line numbering have diverged.
With the mi0bot clone absent, a cite reading `// From mi0bot
console.cs:25081` was resolved against ramdor's console.cs, harvested
whatever author tags sat near that line, and reported our port as
missing them.

This produced 5 confident FAILs on clean main naming //N1GP and
//DH1KLM, all spurious. A checker that fabricates GPL-attribution
violations is worse than one that reports it cannot check, so the
fallback is removed and a missing clone warns.

## freedv-gui cites were invisible

Roughly 68 cites written `// From freedv-gui <path>:<lines>` matched no
detector: the ramdor pattern wants `From Thetis` or a bare `Source:`.
They were never counted, never resolved and never warned about, while
`load_corpus()` was already merging the freedv-gui author corpus into
the tag list, so those contributor tags were loaded and checked against
nothing. CI never cloned freedv-gui either.

Added a detector and resolver ordered ahead of the ramdor catch-all
(exactly as deskhpsdr is), and pinned freedv-gui in CI at
77e793a37e0ecf6cfa99b7cf21a410fd154e9d94 to match the `[@77e793a]`
stamps from Phase 3J-2 and 3R.

## Three cites corrected

Exposed once the checker actually resolved clones. Comment-only.

- `TciServer.cpp` used `From Thetis` grammar naming `RxChannel.cpp`.
  Thetis has no such file; it is NereusSDR-original. Claiming upstream
  provenance for our own code is the inverse of what this tooling
  guards against.
- `AddCustomRadioDialog.cpp` cited `Designer.cs:64`. That is a filename
  suffix, not a file, and the line was wrong. Real source is
  `frmAddCustomRadio.Designer.cs:105`.
- `P1RadioConnection.cpp` used cite grammar for an internal
  cross-reference, and its line number had drifted.

## Verification

- Coverage: 3158 to 3225 cites scanned.
- Mutation test: deleting a real `//MI0BOT` tag inside a cite window
  fails with exit 1 through the pre-commit hook; restoring it passes.
- With all four clones present the run is clean: zero failures, zero
  upstream-not-found warnings.
- Full build green.

No tolerance widened, no skip markers added.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -101,7 +101,8 @@ jobs:
           path: |
             ${{ github.workspace }}/../Thetis
             ${{ github.workspace }}/../mi0bot-Thetis
-          key: thetis-clones-501e3f5-c26a8a4-v1
+            ${{ github.workspace }}/../freedv-gui
+          key: thetis-clones-501e3f5-c26a8a4-77e793a-v2
 
       - name: Clone Thetis + mi0bot-Thetis upstream (cache miss only)
         if: steps.thetis-cache.outputs.cache-hit != 'true'
@@ -125,6 +126,14 @@ jobs:
         # making the verifier flag four phantom missing tags on the
         # untouched HermesLite cite block we ported from c26a8a4.
         #
+        # drowe67/freedv-gui pinned to 77e793a to match every
+        # `[@77e793a]` inline cite stamp from Phase 3J-2 + 3R. Added
+        # 2026-07-28: the ~20 `// From freedv-gui <path>` cites in
+        # FreeDVReporterClient / PskReporterClient matched no detector
+        # in verify-inline-tag-preservation.py, so they were never
+        # counted and never verified. With the detector added, they need
+        # a clone here or they only ever report upstream-not-found.
+        #
         # Shallow fetch + checkout of the pinned SHA avoids cloning the
         # full git history. GitHub's git server allows `fetch --depth=1
         # <sha>` because uploadpack.allowReachableSHA1InWant is enabled.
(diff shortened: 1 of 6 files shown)


boydsoftprez:
Fixed in `10f78ed3`, and the finding turned out to have a second bug hiding behind it.

**The finding as reported.** Correct, and not freedv-only. The line-spec regex had no `+` in its separator class, so a cite naming disjoint locations matched only the first one and the rest were dropped without a word. The cite still counted as checked, so a contributor tag in a discarded range could be deleted and this gate would stay green. 11 cites in the tree use the form, across Thetis, mi0bot and freedv-gui, so all four detectors now share a single `LINE_SPEC` constant rather than four copies of the same regex.

**What it was hiding.** Once the later locations stopped being discarded, the audit went from 0 failures to 9, all on one cite: `DisplaySetupPages.cpp:907`, which cites `display.cs:2304-2337 + 5400-5448 + 5763`. Those 9 were false positives. `extract_tags_from_region()` scanned the contiguous `min..max` of the parsed line numbers instead of a window around each location, despite its docstring claiming the latter. With the components no longer thrown away, that cite swept lines 2299 to 5768 of `display.cs`, roughly 3,500 lines, and demanded nine `//MW0LGE` tags from upstream code the port never touched. `parse_lines_spans()` now returns disjoint `(lo, hi)` spans and each is padded and scanned on its own.

**Evidence.** 3225 cites across 1204 files scan clean, gate exits 0. A mutation test covers both directions: a tag planted at line 100 of a synthetic upstream cited as `10+100` is missed by the old parse and required by the new one, and a tag planted at line 55, between the two spans, is still not required, so the bridging behaviour cannot come back unnoticed.

One note for the record, not a code change: the pre-commit hook still printed the old "searched ../Thetis, ../../Thetis, ../../../Thetis" message on my machine. That is `core.hooksPath` pointing at an absolute path inside the main worktree, which is parked on another branch, so the running hook is that checkout's copy no matter which worktree commits. The ancestor walk in this PR is correct; it just is not the copy git executes until this lands.

J.J. Boyd ~ KG4VCF

boydsoftprez:
@codex review

PR #313, opened 2026-08-02
feat(tnf): tunable notch filter

boydsoftprez:
## What this adds

Ctrl + right-click any carrier on the panadapter and a narrow slot is cut out of the received audio. Drag to move it, drag an edge or roll the wheel to resize, right-click to bypass or remove. A `+TNF` button places one where you are listening. Settings > DSP > TNF gives a table with exact numbers. Turn on Visual Notch and the waterfall shows the carrier drop instead of lying to you. Notches persist across restarts and are shared by every receiver, since they are stored as real on-air frequencies.

The master switch ships **off**, matching Thetis. Because that means your first notch does nothing until you find the switch, the status-bar TNF light goes amber whenever notches exist and are being bypassed.

## Sources

WDSP `nbp.c` for the DSP, Thetis v2.10.3.15 for behaviour and every constant, AetherSDR `@c6481cbf` for the Qt6 shape and marker rendering. `NotchModel` is a three-upstream port carrying all three headers verbatim.

Notch **depth** and **permanent** notches could not be ported: WDSP's NBP has no depth parameter, and Thetis's own menu handlers for it are empty stubs. Thetis's **CW-pitch correction** is deliberately excluded, because it compensates for Thetis folding pitch into the DDS frequency and NereusSDR does not.

## Bugs found on the bench that the test suite did not catch

All four were invisible from the UI. The marker drew over the carrier, WDSP genuinely held the notch, every flag read healthy, and the audio was untouched.

1. **Notches installed into unbound pool channels.** At connect the pool opens `maxSlices` channels but only one slice exists, so notches were pushed into channels with no resolvable RF origin, anchored to zero. Opening a second pan later inherited a poisoned channel and its notches did nothing until a retune. Notch data and origin are now all-or-nothing.
2. **`tunefreq` and `shift` written from two different stream centres.** WDSP maps a notch with `offset = tunefreq + shift` (`nbp.c:192`). After a CTUN pan drag the two terms described different origins and notches landed up to 16 kHz off. Both now have a single writer.
3. **Width presets promised what the DSP would override.** 50 Hz is below the achievable minimum at `nc` 4096, and WDSP's auto-increase silently widened it to 100 Hz. Presets now follow `minNotchWidthHz()` and disable what cannot be realised.
4. **Roughly 100 filter redesigns per second during a drag.** One full `UpdateNBPFilters` per mouse-move per channel, from the GUI thread. Coalesced to 20 Hz with an immediate flush on release.

Plus a segfault caught by the suite: three notch readbacks dereferenced WDSP state on out-of-range channels, where the plain setters only scribble.

## Testing

598/598 pass. 14 new test executables.

**Bench: 5 of 34 rows, ANAN-G2E only.** Matrix at `docs/architecture/2026-07-28-tnf-verification/README.md`.

Row 17 (cold notch on a second pan) is the row least likely to generalise from that pass: the pool defect depends on the channel pool opening more channels than there are slices, and `maxSlices` is 5 on G2 and G2E, 2 on HermesII, and 1 on HL2 where it cannot occur at all. CW mode (row 31) has never been listened to.

Draft until at least rows 14, 17 and 29 are ticked.

## Design

- `docs/architecture/2026-07-28-tunable-notch-filter-design.md`
- `docs/architecture/2026-07-28-tnf-spec-adversarial-review.md`
- `docs/architecture/2026-07-29-tunable-notch-filter-plan.md`
- `docs/architecture/2026-07-28-tnf-verification/README.md`

J.J. Boyd ~ KG4VCF


boydsoftprez:
@codex review

boydsoftprez:
Both findings addressed in `2b5cb9e0`. Both were real.

**Notch capacity.** `NotchModel::addNotch` now refuses at 1024 and emits `notchAddRejected`, and `restoreFromSettings` truncates with a warning rather than repopulating exactly the state `addNotch` refuses to create. The design doc had noted the cap as "unreachable in practice given the 10 Hz dedupe" and then never enforced it anywhere, which is the weaker half of the finding: restore was a second way in that the dedupe does not guard.

**Minimum width on the creation path.** This one is the same defect found on the bench on 2026-08-02, where 50 Hz was selectable in the width presets while WDSP silently widened it to 100. That was fixed in the presets and not swept for siblings, and this creation path is the sibling. At nc 1024 the minimum is 400 Hz and at nc 2048 it is 200 Hz, both above the Thetis 100/200 Hz defaults, so the marker, context menu and settings table would all have shown a width the DSP was not applying. The create path now clamps to `RxChannel::minNotchWidthHz()` exactly as the presets do.

Two regression tests added: filling to capacity and asserting the next add is refused with a signal and no mutation, and a settings file carrying more than the cap restoring truncated.

Worth flagging for anyone reading later: this branch is bench-verified on 5 of 34 matrix rows, ANAN-G2E only. Row 17 in `docs/architecture/2026-07-28-tnf-verification/README.md` is the one least likely to generalise, since the defect it covers depends on `maxSlices` and cannot occur at all on HL2.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Three of the four addressed in `0c1af8fc`. The fourth was a stale re-report: the capacity guard from `2b5cb9e0` is in place and correct at `NotchModel.cpp:340`, refusing before the append and emitting `notchAddRejected`.

The root problem was that there were **three** separate add routes with three different behaviours:

| route | before |
| --- | --- |
| panadapter gesture | clamped, but against `activeSlice()` |
| `+TNF` button | no clamp, and no DIG term in the centre |
| settings Add button | no clamp |

All three now go through `RadioModel::addNotchForSlice(slice, centreHz, widthHz)`, which resolves the minimum realisable width from the channel belonging to the slice it is handed. `grep` confirms no bare `notchModel()->addNotch()` remains under `src/gui/`.

**On resolving from `activeSlice()`.** That was its own bug and a worse one: clicking a pan activates it in `PanadapterStack` without necessarily changing the active slice, so two pans on different filter sizes clamped against each other. The pan identity is now bound at the connect site, where it is known, rather than added to a `SpectrumWidget` signal that has no way to know its own pan id.

**On the DIG offset.** `composedShiftHz` feeds WDSP an origin including the DIG click-tune offset while `effectiveRxFrequency()` is VFO+RIT only, so in DIGU/DIGL the `+TNF` notch was displaced by exactly that offset. New `SliceModel::demodulatedRxFrequency()` carries the complete origin; `effectiveRxFrequency()` is unchanged because every other caller depends on it excluding DIG.

Worth stating plainly, because it is the actual lesson here: this is the third time this defect class has surfaced and the third time it was fixed only where it was pointed out. It was found in the width presets on the bench, then in the creation path by the first review pass, then in the two remaining routes by this one. Collapsing to a single creation route is what ends it, not another spot fix.

New test asserts an add on slice B clamps to B's minimum while slice A is active with a different filter size. 598/598 pass.

@codex review

J.J. Boyd ~ KG4VCF

PR #314, opened 2026-08-02
build: build the application as a shared library

boydsoftprez:
## What

Builds the application as one shared library (`NereusSDRLib`, formerly the `NereusSDRObjs` OBJECT library) so the 598 test executables link it dynamically instead of each embedding a private copy of the whole app.

Written 2026-07-31 on `build/shared-app-library` and never pushed. Rebased onto main across 405 commits, re-measured, and corrected where its own numbers were wrong.

## Why

Apple Silicon (6P + 12E), RelWithDebInfo, ninja, ccache warm, `ctest -j10`:

| | OBJECT | SHARED | |
| --- | --- | --- | --- |
| Touch one `src/core` file, rebuild `all_tests` | 5 min 10 s | **24.3 s** | 12.8x |
| ninja steps for that rebuild | 2,914 | 602 | 4.8x fewer |
| `build/tests` on disk | 24 GB | **2.0 GB** | 12x |
| `tst_smoke` | 41,778,776 B | **88,760 B** | 470x |
| `tst_adif_parser` | 41,799,232 B | **144,464 B** | 289x |
| Full suite | | **598/598 in 109.2 s** | |

Load average was 14.75 entering the OBJECT run against 3.75 entering the SHARED one, so 12.8x is not exact. It is well outside what that gap accounts for.

## The mechanism is not the one the design docs argue

Both design docs credit binary size and the macOS first-run malware scan. Those are real and they dominate the cold suite run. They do not explain the incremental rebuild, which is the case that actually governs the edit-verify loop.

As an OBJECT library the application objects are direct sources of every test target, so touching one `src/core` file invalidates all 598 targets' AUTOMOC before any link is reached:

```
597  timestamp               AUTOMOC, per test target
598  mocs_compilation.cpp.o  moc recompile, per test target
609  .cpp.o                  test TU recompile
598  links
```

As a shared library it is a link-time dependency only. AUTOMOC never fires, no test TU recompiles, and the same edit costs 602 steps.

## Verified locally

598/598 tests pass. `-DNEREUSSDR_ENABLE_LTO=ON` configures clean. The macOS bundle stages the dylib into `Contents/Frameworks` ahead of `codesign --deep`, so the signature validates.

## Not verified, needs CI and a release dry run

Linux and Windows test builds. Only the Linux job sets `NEREUS_BUILD_TESTS=ON`, so `WINDOWS_EXPORT_ALL_SYMBOLS` is unexercised against the suite and the data-symbol limitation stays theoretical until it is.

All three release artifacts launching from a clean environment. v0.5.0 shipped a Windows build that could not start because `rade.dll` was missing, and this adds a second shared library to that surface. `release.yml` gains presence checks on all three platforms plus an `otool -L` guard that fails the macOS build if the dylib still references Qt from outside the bundle.

## Open item

With LTO enabled the app links `NereusSDRObjs_LTO` statically and never loads the dylib, yet the dylib is still built and staged. An LTO release would ship roughly 24 MB nothing opens. Harmless at the current default of LTO off, and left as a follow-up rather than redesigned mid-rebase.

## Also in this branch

Fixes a pre-existing flake in `tst_tx_mic_source` that asserted a no-drop guarantee the ring never made. It overwrites on overrun by design, per Thetis `cmbuffs.c:108-109 [v2.10.3.13]`. The fix also moves its assertions off the consumer thread, where `QVERIFY` was silently truncating the loop instead of failing the test.

Corrects the phase 1 design's own measurements, wrong three times across its history, and replaces the "~32 min" suite figure in CLAUDE.md that every agent reads at session start.

J.J. Boyd ~ KG4VCF

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -319,7 +319,7 @@ jobs:
         # -DNEREUS_GPU_SPECTRUM=OFF on Linux.
         #
         # mold (the apt mold package) is the fast linker. With 451 test
-        # executables linking against NereusSDRObjs + Qt6 the default
+        # executables linking against NereusSDRLib + Qt6 the default
         # ld linker dominated Linux build time at ~33 minutes. The
         # Configure (Linux) step below selects mold via -fuse-ld=mold.
         if: runner.os == 'Linux'
@@ -340,7 +340,7 @@ jobs:
       # SDK / .NET / GHC / unused hosted-tool-cache entries.  Phase
       # 3J-1 closeout pushed the total link-time on-disk footprint
       # past the remaining free budget: 451 test executables each
-      # linking against NereusSDRObjs (~700 MB stripped, much larger
+      # linking against NereusSDRLib (~700 MB stripped, much larger
       # with debug info) plus the vendored RADE neural-codec tree.
       # Run #25835957362 died mid-link with "No space left on
       # device" and the log blob got truncated before upload finished.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -612,11 +612,20 @@ jobs:
         if: github.event_name == 'pull_request' && (matrix.shard == 0 || matrix.shard == '')
         id: locate
         shell: bash
+        # The app is a thin executable over the NereusSDRLib shared library,
+        # so the library has to travel with it or the artifact cannot start.
+        # macOS needs no extra entry: the .app already carries its own copy in
+        # Contents/Frameworks, staged by the CMake POST_BUILD step.
+        #
+        # These artifacts were never fully standalone (they still expect Qt,
+        # librade and friends from the build environment). The point here is
+        # only to keep them exactly as usable as they were before the app
+        # became a shared library.
         run: |
           case "${{ runner.os }}" in
-            Linux)   echo "binary=build/NereusSDR" >> "$GITHUB_OUTPUT" ;;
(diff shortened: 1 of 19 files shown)


boydsoftprez:
Addressed in a5bfbc0c, which landed after the commit this review covers (3a119a8a).

The PIC finding is correct, and it is fixed the way you suggest: `CMAKE_POSITION_INDEPENDENT_CODE` globally rather than per-archive, so the next vendored library does not reopen the same hole. `r8brain`, `rnnoise` and `libspecbleach` already set the property themselves; `wdsp_static`, `portaudio_static` and the rade/opus/lpcnet targets did not.

There was a second failure in the same class that this review did not reach, because it only surfaces on Windows. `WINDOWS_EXPORT_ALL_SYMBOLS` is implemented only for MS-compatible tools per CMake's documentation, and the Windows job builds with MinGW, where the property is a silent no-op. GNU ld falls back to exporting everything only while nothing has been exported explicitly, and zlib is built as a DLL in that job carrying `__declspec(dllexport)`, which switched the fallback off. The DLL linked clean and `libNereusSDRLib.dll.a` came out without a single NereusSDR symbol, so every reference from `main.cpp` was undefined. Fixed by requesting `--export-all-symbols` outright on the MinGW path and keeping the property for MSVC.

Both are consequences of going SHARED that the macOS job cannot catch, since arm64 Mach-O is always position-independent and has no export-list concept.

All checks are green now, including the full suite running under dynamic linking on the four Linux shards.

@codex review

J.J. Boyd ~ KG4VCF

PR #316, opened 2026-08-03
Bottom banner cleanup + AetherSDR-shaped pan menu

boydsoftprez:
The bottom banner needed roughly 1740 px to render its 33 widgets against a
1512-logical-point display, so it overlapped itself whenever the app was open,
not just at narrow widths.

Three responsive systems were competing for the same width. `RxDashboard` ran
its own ladder, `reapplyRightStripDropPriority` ran a second, and the left
section had no policy at all, which is where the visible collision happened.
Each measured itself, changed itself, then measured again, so two carried
hysteresis patches to damp the loop they created.

**Part A** cuts the requirement to about 1286 px and replaces all three with
one authority. Widths are cached once, folding is a single pass over a static
table, and one window width always yields exactly one layout, so flicker is
impossible rather than damped. Both hysteresis hacks are deleted.

- RX pills densified and bound to the **active** slice. They were pinned to
  slice 0 and never rebound, so since multi-pan shipped an operator on Slice B
  saw Slice A's DSP state presented as current.
- Model and firmware fold into `StationBlock`; the duplicate left-side pair is
  gone.
- PA telemetry merges with CPU into one tile, since those rows were already
  mutually exclusive per board.
- UTC clock moves to the title bar, freeing the bottom-right corner for alarms.
- Safety indicators get permanently reserved slots. Previously the ADC-overload
  badge was inserted between PA and TX, so **the TX indicator slid sideways
  exactly when an alarm fired.**

**Part B** replaces the `+PAN` text pill with AetherSDR's drawn spectrum icon
opening a painted nine-layout thumbnail grid, ported at `@c6481cb`. Layouts the
connected radio cannot host are hidden with a footer naming the board, because
we allocate DDCs locally and can say why. Four new layouts added to
`PanadapterStack`. Add-slice and float move off an active-pan lookup onto each
pan's own right-click menu.

**Merged with main**, including the tunable notch filter (#313). TNF was
reclassified in the process: this branch had treated it as an inert placeholder
that folds last, but #313 made it a live toggle that turns amber when notches
exist and are bypassed, so it now folds after even the stubs.

**Bench fixes from a live ANAN-G2E.** Band-stack ordering restored; clock moved
clear of the volume slider; PA supply voltage fixed (a pre-existing race where
the first sample arrived ~19 ms before the listener could bind, and dedup
suppressed every identical one after); TX inhibit moved onto the TX badge as a
prohibition symbol plus a toast, replacing a meaningless `INH` pill that dimmed
into an empty red outline; and `OVERLOAD` restored in full in a slot sized to
it after it rendered clipped as `/ERLO/`.

**Testing.** 606 tests, 603 pass. Three failures (`tst_cty_dat_parser`,
`tst_adif_parser`, `tst_dxcc_color_provider`) are pre-existing `__FILE__`
fixture-path issues, verified against the base commit and chipped separately.
All 11 CI compliance gates verified locally.

**Not verified:** Linux and Windows builds, and the seven bench rows in the
design doc's section 9.2. Matrix drafted at
`docs/architecture/2026-08-02-bottom-banner-and-pan-menu-verification/README.md`.

Reviewer note: six defects reached the bench that 606 tests and a twelve-finding
audit missed, and four share a root cause, labels sized to fit boxes already
chosen rather than boxes sized to content. Each now has a test that fails on
reversion, but the pattern is worth knowing when reading the diff.

Design: `docs/architecture/2026-08-02-bottom-banner-and-pan-menu-design.md`
Plan: `docs/architecture/2026-08-02-bottom-banner-and-pan-menu-plan.md`


boydsoftprez:
Both P2 findings fixed in `107a8991`.

**Separator width on refresh** — fixed in the controller rather than at the call sites. `addItem` now records the separator overhead separately, and `setNaturalWidth` re-adds it, so a caller passes only its own widget width (the sole number it can reasonably know) and cannot drop the overhead. That also covers the TCI indicator you flagged and any future item registered with a separator.

You were right about the consequence: with it dropped, the system tile shed its separator on the first CPU tick and the fold plan undercounted the bar. That is precisely the stale-width class this controller exists to eliminate, reintroduced by the fix wave meant to close it.

Added `refreshingWidthKeepsTheSeparatorOverhead`, and verified it has teeth by reverting the fix and watching it fail with the original symptom (`separator overhead was lost, so the bar was undercounted`).

**Cached voltage race** — both members are `std::atomic<float>` now, with `store`/`load` at relaxed ordering. Relaxed is sufficient here: they are independent scalars guarding no other state, and no ordering relationship with surrounding writes is implied. This matches the cross-thread rule in CLAUDE.md, which I had followed in the writing direction and missed in the reading one when I added the Connected-state priming read earlier today.

Local: 7/7 on the affected suites, app builds clean, all 11 compliance gates pass.

@codex review

boydsoftprez:
Both findings from the `107a8991` pass confirmed as real. One fixed here, one deferred with reasoning.

**Overflow chip charged before anything folded** (`ChromeBarController::buildTable`) fixed in `e0e2e3c5`.

Confirmed exactly as described. The chip was charged at rung 0, where by definition nothing has folded and the chip is not on screen. That made the bar ask whether everything fits alongside the chip that reports things did not fit, and the answer decided whether the chip existed. Between "everything fits" and "everything fits plus the chip" there was a band of widths reachable from below but not leavable: rung 0 got rejected over a chip only rung 1 and up puts on screen, so an item stayed folded with room to spare and the chip stayed up naming it. Arriving at the same pixel width from a wider window gave the correct fully-unfolded bar. Two layouts for one width is precisely the history-dependence this controller replaced three systems to remove.

Fixed in the pure math rather than the controller so it is testable without a GUI. `ChromeFoldEntry` gains `onlyWhenFolded`; `requiredWidth` skips such an entry when `foldThroughRung == 0`. Registration goes through a new `addOverflowChip()` rather than a defaulted bool on `addItem`, since exactly one item in the bar is the ladder own output surface and a bare `true` at the call site would not say which.

The chip is still charged from rung 1 up, which is the reason it is registered at all: once it is on screen, a fold step ignoring it frees less than the plan believes.

Evidence: verified by breaking it in both directions, not by reasoning. Removing the skip fails 4 of the 6 new tests. Making the skip unconditional (chip never charged) fails the other 2. Full suite 603/606; the 3 failures are pre-existing relative-fixture-path failures in `tst_cty_dat_parser` / `tst_adif_parser` / `tst_dxcc_color_provider` that pass in CI and are tracked separately.

**Restored layout not clamped to the connected radio** (`PanadapterStack.cpp:217`) confirmed, not fixed here.

Reproduced by reading the path: the `PanLayoutId` restore block applies the saved layout before any radio is connected, which is deliberate and documented, and `populateEmptyPans()` on `Connected` then asks for a slice on every empty pan with no capability check between them. `SliceStreamAllocator::placeSlice` rejects the surplus own-stream requests and the extra pans stay empty. In-session radio switching has the same hole.

Worth noting this is not introduced by this PR, though the PR widens it by adding the 5-pan `3h2` layout. The dialog side is already correct here: `showPanLayoutDialog()` gates on `qMin(maxSlices, userDdcCount)`, so `3h2` is never offered on an HL2. A layout persisted from a different radio bypasses the dialog entirely, which is the gap.

Deferred deliberately rather than patched in a merge: the fix needs a decision on what replaces an oversized layout (largest that fits, fall back to single, or keep the pans and mark them unfillable), and whether the clamp rewrites the persisted `PanLayoutId` or applies in-session only. Rewriting would mean plugging in an HL2 once permanently loses the G2 arrangement. That is a UX call, not a mechanical one, so it gets its own change with the reasoning visible. Tracked as follow-up work off main.

@codex review

J.J. Boyd ~ KG4VCF

PR #317, opened 2026-08-05
Phase 3M-5: TX display revive — the transmit panadapter shows the transmit signal

boydsoftprez:
Revives the May 2026 `claude/tx-display` work onto current main, then closes five defects found on an ANAN-7000DLE bench on 2026-08-05.

The branch had been parked mid-Phase-3 since 10 May and was 1085 commits behind. `claude/tx-display` is untouched as a reference.

## What this gives you

The panadapter source-switches to the WDSP TX analyzer on the MOX edge, fed by the `sip1` siphon at `TXA.c:586` — after the equaliser, leveler, CFC, CESSB, compressor, ALC, the modulators and `xgen`, but before PureSignal predistortion. So the TUNE tone and the two-tone generator are both visible, and what you see is the intended transmit signal rather than antenna readback.

## The five defects, each of which masked the next

**1. Wrong window.** The pan kept its receive span while the analyzer covered 96 kHz, `visibleBinRange` clamped, and the renderer stretched one across the other. The trace sat at the wrong dial frequency while the RF was correct.

Fixed to Thetis's fixed ±4 kHz transmit window (`display.cs:1284-1295` + `:4585-4590`).

**2. Two waterfall writers.** The ticker's "ALWAYS push on tick" was true when it was the only writer. During transmit the TX analyzer pushes rows at its own rate while the ticker re-pushed the last cached receive row, painting the waterfall in alternating bands.

**3. No span clip.** Thetis derives analyzer span and display window from one zoom state (`specHPSDR.cs:565-584`), so its pixels always cover exactly what is on screen. Ours ran full-span and sliced 8 kHz out of 96 kHz afterwards: ~100 real points upsampled twelvefold, which is a broad blob no window setting can sharpen.

**4. `bf_sz` carried the FFT size.** Thetis passes `fft_size` for `sz` and `blocksize` for `bf_sz` (`specHPSDR.cs:624-642`); we passed the FFT size for both. `Spectrum0()` takes no length argument, so WDSP read 32768 samples out of a 2048-sample buffer every frame and transformed whatever followed it in memory. **This is why no window, detector or averaging setting appeared to do anything** — they were being applied to garbage.

Zeroing the symmetric clip belongs with it: WDSP subtracts the span clips from a span already reduced by `2*clp` (`analyzer.c:1283`). Thetis says so plainly in `CalcSpectrum`: *"no need for any symmetrical clipping; int sclip = 0"*.

**5. No transmit grid, and no way to set one.** The pan inherited the receive dB range — chosen for signals near the noise floor — to display something at about −9 dBm. Thetis swaps the whole graticule on the MOX edge (`display.cs:1887-1905`, +20 / −80).

The operator could not work around it either: the TX Grid Scale group in Setup is still a placeholder, so there were no controls, and the rise edge re-imposed defaults every transmission while the fall edge restored the receive grid over the top. A drag made during transmit was discarded twice before the next one. The grid is now seeded from Thetis's constants, then replaced by whatever the operator leaves on screen at un-key and persisted, so the dBm strip is the control.

## Multi-pan correctness

The May work predates Phase 3F. The MOX lambda used `m_spectrumWidget` / `m_fftEngine`, neither of which exists, and disconnected a signal that is now a router lookup. It resolves through `RadioModel::txBoundSlice()` now, so the transmit trace lands on the pan that is actually transmitting — deliberately not `spectrumForSlice()`, whose `activeSpectrumWidget()` fallback would paint transmit onto an unrelated pan.

Fixes a latent bug in passing: `setMoxOverlay` was bound to `activeSpectrumWidget()` once at construction, so after multi-pan landed the overlay kept going to whichever pan was active at startup.

## Sourcing warning worth reading

`UpdateTXDisplayVars` and `CalcSpectrum` govern the SPECTRUM / HISTOGRAM / SPECTRASCOPE displays, **not** the panadapter — `UpdateTXDisplayVars` returns immediately for PANADAPTER/PANAFALL (`console.cs:8026-8029`). `TxAnalyzer.cpp`'s file header already warned about this after the same mistake in May; it was made again on 2026-08-04 and caught at the bench. The warning is now also in the CLAUDE.md phase row.

## Known open

A residual skirt ~35 dB down at 66 Hz from the peak, far worse than Blackman-Harris 4T should give. It is present in WDSP's raw `GetPixels` output, so it is upstream of everything here. Ruled out: display range, resolution, window type, tone magnitude (identical to Thetis's `MAX_TONE_MAG`) and our render path. Next steps are an A/B against Thetis on the same radio, or running the receive stream through both `FFTEngine` and the WDSP analyzer.

Row 15 of the bench matrix, with the measured numbers.

## Verification

- Full suite **607/610**. The three failures are pre-existing relative-fixture-path ones (`tst_cty_dat_parser`, `tst_adif_parser`, `tst_dxcc_color_provider`) that pass in CI.
- All four attribution gates pass across the branch.
- Bench matrix at `docs/architecture/tx-display-verification/README.md`: rows 1-8 verified on the 7000DLE, 9-14 pending, 15 open.

## Also carried

3M-5b TX waterfall colormap and 3M-5c custom gradient picker, both bench-passed in May. Phases 4-6 not started; Phase 4 is partly obviated since the graticule now switches and persists on its own.

J.J. Boyd ~ KG4VCF


boydsoftprez:
@codex review

boydsoftprez:
Both P2 findings fixed in `915c1c38`. Both were real.

**TX-bound slice for the carrier.** Exactly right, and it is the case the arbiter exists to support. The rise edge resolved the target pan from `txBoundSlice()` and then centred the bin mapping on `activeSlice()`. Same object in the common case; they diverge precisely when TX has been handed to a slice the operator is not looking at, and then the analyzer is fed by the transmitter while the window sits at a receive-focused frequency. Trace renders in the wrong place on the right pan. `carrierHz` now comes from the already-resolved `txSlice`.

**Receive scrollback.** Also right, and it broke a requirement stated outright in the 2026-05-08 handoff doc: *"on MOX-down, RX1 spectrum returns to the panadapter without losing waterfall scrollback continuity."* `setFrequencyRange` treats any bandwidth change as a large shift and clears history, which the MOX edge triggers twice per transmission.

Worth noting why the obvious fix was not available: the original code preserved scrollback by deliberately never touching centre or bandwidth, and that is exactly what made the frequency axis lie (the analyzer covered 96 kHz, the pan kept its receive span, and the renderer stretched one across the other). So this could not be reverted, only replaced.

New `setDisplayWindowPreservingHistory` changes the window with no clear and no reprojection. The absence of reprojection is deliberate rather than an omission: transmit rows and receive rows describe different windows, so mapping one onto the other and back loses information in both directions. Leaving the plane untouched is what actually preserves the receive history.

Evidence: 91/91 across the spectrum, pan, waterfall, fft, mox and tx suites. The scrollback behaviour itself is bench-observable rather than unit-testable (it needs a live waterfall across a real key-up), so it is row 10 of `docs/architecture/tx-display-verification/README.md` rather than a test I can claim green.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Both new P2 findings fixed in `b54de725`. Round 2, and both were real.

**Double detection.** This one is the good catch. `TxAnalyzer` output has already had a detector and averaging applied by WDSP using the TX Display controls, and `updateSpectrumLinear` then ran `applySpectrumDetector` and `m_spectrumAvenger` over it a second time using the *receive* controls. So the TX Display detector and averaging only behaved as labelled while receive happened to sit at Peak / None.

New `updateSpectrumFromTxPixels` takes the dBm pixels straight and writes `m_renderedPixels`. The resample inside it is deliberately not a detector: WDSP already reduced bins to pixels, and with the analyzer clipped to the display window `n_pix` and `displayWidth` are within a few percent, so there is nothing left to decimate. It is the trace-plane twin of `setTxExternalWaterfall`, which already stood the receive path down for the waterfall plane.

It also removes a dBm to linear to dBm round trip that existed purely to satisfy a signature. The old adapter raised each pixel to a power so the widget could take its log again; its own comment called it a throwaway debug adapter.

Worth noting for the record: I considered exactly this cause during the bench session and dismissed it, on the grounds that changing the FFT window visibly altered the trace so the analyzer output must be reaching the screen unmodified. That inference was wrong. Both things were true at once.

**FFTW planning on the GUI thread.** Also correct. `SetAnalyzer` builds `FFTW_PATIENT` plans on first sizing, `TxAnalyzer` is constructed from `buildUI()`, and that runs before any connection has started `WdspEngine::initialize()` and its background wisdom path. Cold launch planned a 32768-point transform synchronously before the connection UI appeared.

Deferred to the first `start()`, which sits behind the MOX edge and therefore behind the connection flow. Setters record state throughout and the accumulated configuration is applied in one go when the analyzer is armed, so nothing is dropped.

**On the test:** `tst_tx_analyzer_settings` failed on the deferral rather than being adjusted around it. Its "every setter reconfigures" assertion is now preceded by an explicit arm, and a new case pins the deferral itself so construction cannot quietly start planning again.

Evidence: 11/11 on `tst_tx_analyzer_settings`, 91/91 across the spectrum, pan, waterfall, fft, mox and tx suites.

The two inline comments still showing at line 0 are the round-1 findings, fixed in `915c1c38`.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Fixed in `57f7d365`. Round 3.

**XIT centring.** Correct, and it is the third variation on one theme across these three rounds: the display has to centre on the number the *transmitter* uses, not the one the operator is looking at.

- round 1: `activeSlice()` rather than the TX-bound slice
- round 3: `slice->frequency()` rather than the XIT-shifted carrier

With XIT on, the transmit chain runs at dial + XIT, so the trace, waterfall and TX filter overlay were all drawn around the RX VFO while the radio transmitted somewhere else.

`carrierHz` now comes from `RadioModel::txFrequencyForSlice`, which already owned that arithmetic and documents its Thetis provenance in place (`console.cs:31845-31891` — the TUNE offsets are applied to the same `tx_freq` afterwards, so TUNE transmits on the XIT-shifted frequency too). That helper was private with only a `ForTest` wrapper exposed; I made the real one public rather than call a test seam from production code, since the TX display is a legitimate second consumer of "where is this slice transmitting".

**What I did not do, stated plainly:** the finding also asks that the centre be kept refreshed if XIT or frequency changes *while keyed*. It is not. Centre, window and grid are all computed once on the MOX rise edge, so a mid-transmission XIT nudge leaves the display where it started until the next key-up.

I have deliberately not special-cased XIT for that. Every value on the rise-edge path behaves the same way, so doing it properly means a live-update path for the whole set rather than one field wired differently from its neighbours — which is how the inconsistencies fixed earlier in this PR arose in the first place. Recorded as row 16 of `docs/architecture/tx-display-verification/README.md` so it is visible rather than absent.

Evidence: 131/131 across the spectrum, pan, waterfall, fft, mox, tx, radio-model and slice suites.

This is the third and final Codex round for this PR per the review budget agreed for it. Remaining known-open items are rows 15 (residual skirt, upstream in WDSP raw output) and 16 above, both documented in the matrix rather than carried silently.

J.J. Boyd ~ KG4VCF

boydsoftprez:
Reopening the review budget for one more round. Two commits landed after `57f7d365`, and neither has been looked at.

**`7bf5bf2e`** is the rest of the 2026-08-05 bench wave. It was written after `e05f436d`, sat uncommitted in the worktree for four days, and is the largest single change on this branch since the Codex rounds. Nine defects, three groups:

*View window and analyzer span disagreeing.* Thetis derives both from one zoom state, ours are separate objects, and every writer of `m_centerHz` / `m_bandwidthHz` had to remember to announce itself. Six defects came from writers that did not: a tone stuck where key-down happened, 8 kHz of bins stretched across a 55 kHz axis, a passband sliding under the cursor mid-drag. `applyViewWindow()` is now the only place those two are written outside `setMoxOverlay`'s swap, and it emits `txViewWindowChanged` so the analyzer re-clips. Asymmetric around the carrier rather than a symmetric half-span, because a symmetric clip is right only while the view is centred and the error grows with the pan offset. The view window also becomes per direction, so a transmit zoom survives to the next transmission.

*The dBm strip.* The drag scaled by `m_dynamicRange / specH`, about 0.13 dB per pixel on a 49 dB window, where Thetis uses a fixed half-dB (`PanDisplay.cs:3688-3699`). Its +20 ceiling was exactly the transmit grid's reference level, so dragging up while keyed started pinned against the stop. The scale labels live in the cached overlay texture and the trace does not, so `update()` alone moved the trace and froze the numbers. And a right-press that macOS turned into a context-menu gesture swallowed the matching release, leaving the drag flag set forever. Right-drag also gains its Thetis meaning of ceiling-only.

*Persistence and one safety gap.* `saveSettings` wrote `DisplayGridMax`/`Min` from the live pair, which is the transmit grid while keyed, so a save landing mid-transmission put transmit values in the receive keys. `setMoxOverlay` no longer schedules a save, because doing so laundered the noise-floor tracker's deliberately unpersisted drift into the keys and receive eroded from -1xx dBm to about -90 over a few transmissions. And click-to-tune is withheld while `m_moxOverlay` is set: panning and zooming during transmit are new in this branch, so a stray click could now retune a keyed transmitter into an antenna and an amplifier.

**`8587813e`** fixes something I found reviewing `e05f436d`. The three transmit-only keys went in as bare globals while every other key in `loadSettings` / `saveSettings` goes through `settingsKey(base, m_panIndex)`. That is not a shared setting, it is a race: every pan reads the one key at startup into its own member and writes that member back on every save, so the next save from any quiet pan puts the startup value back and the transmitting pan's drag is silently undone.

**Evidence, and its limits.** `7bf5bf2e` builds clean and 133/133 pass across the spectrum, pan, waterfall, fft, mox and tx suites. But the nine behaviours in it are bench observations from the 2026-08-05 session, not assertions: they need a live carrier and a real pointer, and I have not been back on the radio since. They are matrix rows, not tests, and I am not claiming them verified. `8587813e` is the exception, and it has a test verified red against the bare-key version before the fix went in.

Worth flagging for this round specifically: `7bf5bf2e` routes every window write through one funnel and touches the MOX edge, the settings path and the click router. That is a lot of surface for code that reached main-line review only now.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
All five fixed in `6eded0f3`. All five were real, and the P1 was the one worth reopening the budget for.

**Every pointer path, not just the one.** Correct, and it is the more embarrassing kind of miss because the guard existed and I checked the wrong thing: I confirmed the release branch was gated and never asked what else emits `frequencyClicked`. Eight things did. A press inside the passband sets `m_draggingVfo` and `mouseMoveEvent` tunes on every move after it, a spot label tunes on press, the wheel tunes on every notch. So the exact hazard the guard was written for, a stray gesture moving a keyed transmitter into an antenna and an amplifier, was still open through three doors.

`requestTune()` is now the only emitter and it carries the gate. The release branch keeps its own `!m_moxOverlay` because `ddcRetuneRequested` sits in the same block, is not routed through `requestTune`, and moves the slice just as surely: its own comment records that MainWindow forwards it to slice frequency. Menu actions go through the gate too, on the grounds that the radio moving while keyed does not care whether the operator meant it.

**Forcing the initial re-clip.** Right, and it is this branch's own design biting back. `applyViewWindow` deliberately emits nothing when nothing moved, which is what stops mouse-move spam, and the rise edge was leaning on it to emit. When the receive view already sat on the carrier at the transmit span there is nothing to move, so the analyzer kept the hard-coded 8 kHz seed while the pan drew the restored wider axis. That is the same defect `7bf5bf2e` was written to fix, reappearing through the door it left open. The handler is now a named lambda called explicitly once after the connect.

**Clamping the view, not just the analyzer.** Also right. I clamped `lo`/`hi` and left the widget showing a window the analyzer cannot fill, so `updateSpectrumFromTxPixels` stretched the clipped bins across the whole axis; and a window driven almost entirely outside the baseband tripped the too-narrow return and left the previous mapping under a new axis.

Of your two suggestions I took the first. Rendering the unavailable portion as empty is defensible, but the transmit display genuinely cannot show more than the siphon's 96 kHz, so stopping the operator at the edge seemed more honest than showing them a permanently dead margin. The clamp is re-entrant by exactly one pass: the push re-emits with the clamped values, the second pass finds nothing to clamp, and `applyViewWindow` stops there.

**Both persistence bounds.** Both right, and both are the same commit failing to finish its own thought. The ref-level loader kept `[-160, +20]` while the new drag clamps to Thetis's `[-200, +200]`, and +20 is precisely the transmit grid's default reference level, so dragging a pinned scale up saved the fix and had it thrown away at next launch. And `saveSettings` protected the receive dBm pair from a keyed save while `DisplayBandwidth`, one line down, still read the live `m_bandwidthHz`.

**Evidence.** Three new cases, all verified red against the pre-fix code before the fixes landed: with the gate removed, the `DisplayBandwidth` write reverted and the bounds narrowed, the run reports 10 passed / 3 failed. 13/13 on `tst_spectrum_widget_mox_overlay`, 133/133 across the spectrum, pan, waterfall, fft, mox and tx suites.

**Stated plainly:** the tune gate is asserted through `requestTune` rather than by synthesising a press, a wheel notch and a spot-label hit. What makes that cover the three paths it does not drive is that they are all routed through the one function, which is a grep and not a test. And none of this has been on a bench since; the analyzer sync and the baseband clamp are both things I would want to watch on a real carrier before calling them confirmed.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Both fixed in `2f00e41b`.

**TX Low Color never reached a pixel.** Right, and it turned out to be a half-finished port rather than a missing feature. Thetis sets the threshold and the colour together in its TX branch, `display.cs:6424-6427 [v2.10.3.15]`:

```
low_threshold  = (float)TXWFAmpMin;
high_threshold = (float)TXWFAmpMax;
cScheme        = _tx_color_scheme;
low_color      = waterfall_low_color_tx;
```

consumed at `:6766-6768` under `if (waterfall_data[i] <= low_threshold)` at `:6764`. NereusSDR ported the first two lines and left the fourth, so `m_txWfLowColor` was settable, persisted, read back on load, and never consulted: a bin at or below `m_txWfLowLevel` produced `adjusted == 0` and fell into the palette's first gradient stop. TX branch only, since receive has no separate low-colour control on this path.

The attribution gate is worth recording here. My first attempt cited the render region as a broad `display.cs:6826-6954` range, and `verify-inline-tag-preservation.py` blocked the commit for three `//MW0LGE` tags inside it that the port does not touch. The gate was right to: a range that wide claims provenance over code I did not port. Narrowing the cite to the four lines actually ported is both the correct attribution and what clears the check.

**The swallowed release.** Also right, and it is the fallback failing at the one job it was added for. `7bf5bf2e` added the buttonless-move termination precisely because macOS can eat the right-button release, and it cleared the flag and stopped: no `dbmRangeChangeRequested`, no `scheduleSettingsSave`. So on exactly that case the operator drags the scale into shape, it holds for the session, and it is gone at next launch, having done nothing wrong. Both fallbacks now finalise the same way `mouseReleaseEvent` does.

**Evidence.** Two new cases, both verified red against the pre-fix code: with the colour branch removed the run reports 13 passed / 1 failed. 14/14 on `tst_spectrum_widget_mox_overlay`, 133/133 across the spectrum, pan, waterfall, fft, mox and tx suites.

**Not claimed as verified:** the colour fix is asserted through `dbmToRgbForTest` rather than against a rendered frame, and the swallowed-release path is macOS window-server behaviour that no unit test here reproduces. Both are right by construction; neither has been on a bench.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
@codex review

PR #318, opened 2026-08-09
Fix floated-pan rendering, splitter grab area, and render the extended pan's wideband wings

boydsoftprez:
Three bench-reported defects, plus the feature behind the third turning out to be unbuilt rather than broken.

## Floated pan stopped rendering

A popped-out pan painted one frame and froze, logging `QRhiWidget: No QRhi` once per display tick. Confirmed on an ANAN-G2E: one `CAMetalLayer` acquired, then 30 warnings a second, waterfall static.

The existing workaround hid and re-showed the *applet*, but hiding a parent does not make a `QRhiWidget` release its render context, so nothing ever told the `SpectrumWidget` its window had changed. Ported AetherSDR's teardown sequence, which fixed the same defect on the same widget class.

The refresh runs on **only** the widget whose top-level actually changed. A first attempt ran it over every docked pan, which killed whichever pan had *not* moved: destroying the `QWindow` of a native child whose parent window never changed tears down a live Metal layer that nothing re-creates, and does it silently with no warning logged.

## Floated pan had no container chrome

It now arrives parented to the console with a title strip carrying the pan name and a Dock button, plus per-pan geometry persisted under `FloatingPan_<panId>_Geometry`. The strip is floating-only. Upstream shows it on docked pans too, but adopting that would put a bar on every docked panadapter, which is a redesign of the docked stack rather than a fix to the float path.

## Splitter between pans was hard to hit

`PanadapterStack` never called `setHandleWidth`, so every splitter fell back to the style metric. A `QSplitter`'s drag target *is* its handle rect, so the target was a couple of physical millimetres of unpainted gap. Now one `makeSplitter()` factory at 8 px, non-collapsible, rather than a setter at each of 11 call sites that the next layout would forget.

**Wants review:** the handle keeps its hairline paint at rest so a docked stack looks unchanged, but hovering lightens it. That hover state is the only visual addition in this PR.

## Extended view: the render half was never written

Sub-Epic F shipped the entire wideband *data* path in May and no paint. `setWidebandBins` stored the bins and said so in a comment; nothing read them back. Four defects sat behind that:

- **Unreachable state.** Both zoom gestures clamped bandwidth to the DDC rate, and `m_bandwidthHz > m_sampleRateHz` is extended mode's only trigger. The toggle enabled a state no gesture could reach.
- **Stretched island.** `visibleBinRange()` clamps to the DDC's bins, so zooming past the DDC rate spread its spectrum across the whole panel instead of its true span, leaving the trace disagreeing with the frequency scale beneath it.
- **Amplitude.** Two derived terms: a coherent-gain reference following the analysis window, and a noise-bandwidth term referring the wideband bin to the DDC bin. The second is what saturated the wings on the bench: both halves show noise, noise power scales with bin width, and 7.5 kHz against 11.7 Hz overstated the wideband by 28 dB.
- **Filter skirt.** The outermost DDC bins are transition band and read near-nothing, showing as dark bands at the island edges. Thetis discards 4% per side on its own display (`specHPSDR.cs:529`); so do we, island geometry following so signals keep their frequencies.

`WidebandFftEngine` also gains a Hann window and 4x zero-padding. It ran unwindowed, so rectangular with -13 dB sidelobes, smearing every strong broadcast carrier. Blackman-Harris 7-term (Thetis's own choice) was tried and **rejected on the bench**: its 2.63-bin ENB widened effective resolution from 7.5 kHz to 19.7 kHz. Upstream affords that because its wideband window is a standalone survey, not something beside a DDC island. The padding is interpolation, not resolution, and the engine publishes `binSpacingHz()` and `noiseBandwidthHz()` separately with a test asserting they stay apart so the finer spacing is never misread later.

True wideband resolution is 133 us of contiguous capture, fixed by the FPGA's 16k FIFO. `Orion.v:1503` carries the upstream TODO to make it selectable.

## Verification

Full suite **604/607**. The three failures are pre-existing and unrelated (`tst_cty_dat_parser`, `tst_adif_parser`, `tst_dxcc_color_provider` resolve a data file from a relative `__FILE__`). Attribution gates green: tag preservation, provenance sync 319/319, Thetis headers 352/352, no new unstamped cites.

Bench-verified on an ANAN-G2E: float renders continuously, chrome and geometry work, extended view engages with WIDE badge and bypass, bypass releases on zoom-in.

**Not bench-verified:** the float/dock round trip after the scope fix, and the wings after the Hann and skirt-clip changes.

## Not in this PR

Two pre-existing defects found while working, both diagnosed, neither touched: the `cty.dat` test path bug above, and a GUI hang on reconnect where the main thread blocks in `AudioEngine::stop()` under `teardownConnection()` (sampled stack captured).

J.J. Boyd ~ KG4VCF


boydsoftprez:
@codex review

boydsoftprez:
Both P2 findings fixed in `d182f6b4`. Both were real, and the second one is the better catch.

**The 4% strips belonged to nobody.** Correct, and the click half is the part that matters. The island is painted `kDdcClipFraction` narrower than the DDC rate at each edge because those bins are decimation-filter skirt, and only the paint knew it: `drawExtendedIslandBounds` marked the boundary at `rate/2` and `mouseReleaseEvent` hit-tested against `rate/2`.

I had spotted the marker offset while reviewing and written it off as sub-pixel at these zooms, which it is. I missed that the same number drives the click router. A marker in the wrong place is cosmetic. A click that retunes the slice when the pixel under it is wideband survey data is a disagreement the operator finds out about only after the radio has moved.

`ddcIslandHalfSpanHz()` is now the single definition and all three consumers read it.

**Detector normalisation in the wing calibration.** Also right, and it is the kind of thing that hides because it is correct in the default. `applySpectrumDetector` is handed `invEnb = 1 / m_fftWindowEnb`, and only cases 2, 3 and 4 apply it; cases 0 and 1 take a max and never touch it. So under Average, Sample or RMS the island pixels have already had the window ENB divided out and their effective reference is the bare bin width, while the wings were still being referred to `binWidthHz() * m_fftWindowEnb`. About 1.8 dB high on Hann, near 5.8 dB on Flat-Top.

The detector is now an argument rather than a member read, for the reason you gave: the trace and the waterfall select independently, so a single member could only ever have been right for one of the two planes. Also hoisted out of the per-pixel loop, where the whole calibration was being recomputed for every wing pixel.

**One thing you did not flag, fixed alongside.** The comment on `widebandFftNormalisationDb` still described a Blackman-Harris 7-term window and its roughly 24 dB coherent gain. BH-7 was tried and rejected on the bench the same day for its 2.63-bin ENB, and the engine ships Hann, where the figure is about 6 dB. The code reads `windowSum()` at runtime so it was never wrong, but the note would have misled whoever touched this next.

**Evidence.** Both fixes have a test case verified red against the old behaviour before the fix went in, rather than written after the fact. `bandwidth_term_follows_the_detector` pushes a frame at Hann's 1.5 ENB and asserts Peak and Average differ by exactly `10*log10(1.5)`; with the branch forced off it reports a 0 dB gap against an expected 1.76. `island_half_span_is_one_number` carries a guard that fails outright if the span ever stops being clipped, so it cannot quietly stop discriminating. 40/40 across the spectrum, pan, waterfall, fft, wideband, extended and float suites.

Not claimed as verified: neither fix has been on a bench since. The island geometry and the calibration are both arithmetic the tests can pin, but the click-routing change is behaviour, and it wants a real pointer on a real extended pan before I would call it confirmed.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
All three fixed in `97a5338b`. All three were real.

**Routing the wings by filter chain.** Right, and the comment I wrote defending it is the tell: it argued that the chain index is the number RadioModel hands `setWidebandEnabled`, which is true and is not the same as being the right number to paint by. `chainForStream` deliberately folds ADC1 onto chain 0 when a board has more ADCs than preselector banks, and `RadioModel.h` spells out ANAN-100D and ANAN-200D by name as the case. The wideband frame goes nowhere near that fold: `widebandSpectrumReady` carries the physical index straight off the wire. So on exactly the boards the fold exists for, an extended pan fed by ADC1 drew ADC0's survey in both wings either side of a perfectly correct DDC island. Unrelated signals at plausible levels, which is the worst way for a display to be wrong.

New `adcForStream` / `sliceAdcIndex` are the unfolded pair. Deliberately separate functions rather than a flag on the existing ones: the fold is right for every filter question, and the two callers should not look alike at a glance.

**The one-pixel island.** Right, and the second half of your finding is the one that matters. The manufactured one-pixel span is a small lie; the frozen display is not. `visibleBinRange` clamps to the terminal bin, the 4% skirt clip then inverts the range, `sliceCount` goes negative and `updateSpectrumLinear` returns before `fillWidebandWings` ever runs. So panning far enough into a wing stopped the survey dead on stale pixels, which reads as a broken panadapter rather than as a pan that has gone where the DDC cannot follow.

No-overlap is now explicit: an empty span from `listenableIslandPixels`, both detector calls skipped, and the wideband plane owning every pixel. Outside extended mode an empty slice still means nothing to draw and returns exactly as before.

Worth recording: `tst_extended_pan_wings` had a case that asserted the defect. `island_clamps_when_the_ddc_leaves_the_window` put the DDC 5 MHz off-screen and ended in `QVERIFY(first <= last)`, so it was pinning the manufactured island as if it were the contract. I replaced it rather than relaxing it. The clamp case keeps a test, with the DDC straddling the window edge where clamping is genuinely correct, and the no-overlap case gets its own.

**Floating geometry on quit.** Right, and it is the ordinary path that was broken: `dockPanadapter` saves and the close box saves, but `dockAllFloatingPans` did not, and that is where the destructor goes. Quitting with a pan popped out is how a session normally ends, so the last move or resize of the day was the one guaranteed to be lost.

**Evidence.** 18/18 on `tst_extended_pan_wings`, including a new `island_is_empty_when_the_ddc_is_off_screen`, and a new `floating_geometry_survives_stack_teardown` that destroys a stack with a pan still floating. 40/40 across the spectrum, pan, waterfall, fft, wideband, extended and float suites.

**Not claimed as verified:** the ADC-routing fix cannot be exercised on the bench I have. It needs a 2-ADC / 1-chain board (ANAN-100D or 200D) with a slice on ADC1 and extended view engaged, and the failure mode is a wing that looks plausible, so it will not announce itself. It is right by construction and by the provenance of the two indices, and that is the strongest claim I can make for it.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
All four fixed in `ca34021b`. One of them is a retraction, so that first.

**I said the floating geometry was fixed. It was not.** `97a5338b` added `saveWindowGeometry()` to `dockAllFloatingPans`, and you are right that it never reaches disk: `MainWindow::closeEvent` calls `AppSettings::instance().save()` and then `QCoreApplication::quit()`, and `~PanadapterStack` runs after that. `AppSettings::setValue` only inserts into an in-memory `QMap` and `~AppSettings` is defaulted, so the teardown save wrote into a map nobody flushed. The behaviour on the operator's disk was identical before and after that commit and I claimed otherwise.

What I got wrong was the shape of the check. I verified the save was on the teardown path and stopped, without asking when that path runs relative to the flush. New `saveFloatingGeometry()` is called from `closeEvent` next to `saveSplitterState`, before the flush. The teardown save stays, because it is still correct for the non-quit paths where a layout change docks everything while the app keeps running.

**The ADC lookup was right and never re-ran.** Correct, and it is the same class of miss as the P1 on the sibling PR: I fixed what a function computes without asking when it is called. A move from ADC0 to ADC1 changes neither the stream index nor, on a one-chain board, the folded chain index, so `statusOverlaySliceProperties` sees nothing move and the refresh never fires. `publishDdcAssignment` now notices when an `m_streamAdc` entry actually changed and emits `streamAdcRoutingChanged`; MainWindow wires it to the refresh. Guarded on a real change, so a re-publish that moves nothing stays silent.

**Both spans that outlived their ceiling.** Both right, and both follow from `maxZoomOutBandwidthHz` being consulted by the two zoom gestures and by nothing else.

Withdrawing permission left the wide span in place with `m_extendedMode` cleared, so the next frame stretched the DDC's bins across the whole panel with no wings: the stretched-island defect reachable through the toggle that is supposed to be the way out of it. And a span above the DDC rate is now a legitimate state, since it is extended mode's only trigger, so `wireSliceToSpectrum`'s hardcoded 768 kHz test and the sample-rate handler's clamp were throwing away a correctly persisted extended zoom on every restart and reconnect. Both now measure against the extended ceiling, which returns the DDC rate when extended view is off and leaves the ordinary case untouched. `wireSliceToSpectrum` also sets the sample rate before reading the ceiling, or it would compute one from the widget's construction default.

**Evidence.** Two new cases verified red against the pre-fix code: `disabling_extended_view_clamps_the_current_span` and `floating_geometry_is_saved_before_the_flush`. 19/19 on `tst_extended_pan_wings`, 22/22 on `tst_panadapter_stack_layouts`, 40/40 across the spectrum, pan, waterfall, fft, wideband, extended and float suites.

**Standing caveat, unchanged:** the ADC-routing fix still cannot be exercised here. It needs a 2-ADC / 1-chain board with a slice on ADC1 and extended view engaged, and its failure mode is a wing that looks plausible.

@codex review

J.J. Boyd ~ KG4VCF

boydsoftprez:
Fixed in `fb239694`. Real, and it is the third copy of the same clamp.

`ca34021b` fixed the startup restore and the sample-rate handler and left `applyStreamWindowToPan` alone, which is the one that runs last: `streamBindingsChanged` triggers `rebuildFftRouting`, whose new-subscription path calls it, and it clamped every span above the stream width straight back down. So the restored extended zoom was accepted at startup and collapsed the moment routing was established. Same collapse for any pan that later gains a subscription.

I could have fixed the site you named. On this branch that has been a reliable way to be back here next round: this is the third clamp, found one per round, and the P1 on the sibling PR was the same shape (one tune path gated out of four). So `setDisplayWindowClamped` is now the shared definition and two of the three sites call it. Non-positive or above-ceiling becomes the ceiling; the ceiling is `maxZoomOutBandwidthHz`, which is the DDC rate whenever extended view is off, so the ordinary case is untouched.

`wireSliceToSpectrum` deliberately does not call it. Its fallback genuinely differs: an out-of-range stored value goes to the full-span 768 kHz default rather than to the ceiling, and its 10 kHz floor applies only to a restored value. It already reads the same ceiling, and the comment now says why it keeps its own shape rather than leaving that looking like the next oversight.

New `clamped_window_keeps_an_extended_span` pins all four behaviours: an extended span passes through, an absurd one clamps to the ceiling, non-positive means the ceiling, and with extended view off the ceiling is the DDC rate again. 20/20 on `tst_extended_pan_wings`, 40/40 across the spectrum, pan, waterfall, fft, wideband, extended and float suites.

Worth saying plainly, since this is round 4 on this PR: the findings have not been noise, but they have been arriving faster than they have been closing, and three of them now have been "you fixed the instance and not the shape". I have flagged that to the maintainer as a question about the PR rather than about any individual finding.

@codex review

J.J. Boyd ~ KG4VCF

